PR #9, opened 2023-11-05
Untangle all of the DriveWire modules; Also separate DWWrite and DWInit which got mushed together

n6il:
@boisy this is one possible version of un-tangling all the DriveWire modules. This one still uses conditional assembly but significantly simplifies the logic and moves all of the code to separate files.

Let me know what you think of doing it this way.

Along the way I found that in `dwio.asm` the code for the `Write` block and `DWinit` blocks got mushed together somehow.  While the code really isn't wrong it just looks like it, so I put another label in there just to separate them.


boisy:
This is much needed and appreciated work. Thanks!

boisy (approved):

strickyak (approved):
Looks good to me!

PR #39, opened 2024-01-03
Backport comments & symbols & H6309 L2->L1 for RBF

strickyak:
The idea at this stage is to reduce the diffs as much as possible in these two versions of rbf.asm:
            level1/modules/rbf.asm
            level2/modules/rbf.asm
WITHOUT making any changes to their binary output. These md5sums are preserved by this change:
  nitros9$ md5sum $(find level* -name rbf.mn)
  7ab6d3a0590edbe76b1528fd861876dd  level1/coco1/modules/rbf.mn
  b1553307099d67b0c5d8a0b80c4fd2e9  level1/coco1_6309/modules/rbf.mn
  3c78bca12d14f22dbfd9e2ab1fc21ce9  level2/coco3/modules/rbf.mn
  ac01408b7802f826737a36fcbbe99777  level2/coco3_6309/modules/rbf.mn

The diffs to level1/modules/rbf.asm are enormous, and will be best understood by using something like vimdiff to compare level1/modules/rbf.asm to level2/modules/rbf.asm.

Also H6309 changes were backported from level2 to level1 when they looked safe.  They are conditionally compiled as UNTESTED_H6309, so they will never be used in ordinary builds, until we can give them a good testing first.

### DIFF
--- a/level2/modules/rbf.asm
+++ b/level2/modules/rbf.asm
@@ -1,7 +1,21 @@
 ********************************************************************
 * RBF - Random Block File Manager
 *
-* $Id$
+* nitros9/level2/modules/rbf.asm
+*
+* 2024-01-02 by strick:
+* This version of level2/modules/rbf.asm has been altered slightly
+* to minimize diffs with the level1 version of rbf.asm.
+*
+* So far, we've been careful that level2 rbf assembles to exactly
+* the same binary as before, both for the M6809 and the H6309 case.
+*
+* As you edit either level1 or level2 rbf.asm, please minimize
+* diffs between the two, in case we decide to do the work to
+* unite them.  If possible, edit them in parallel with something
+* like vimdiff.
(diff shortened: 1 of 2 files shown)


level2/modules/rbf.asm
```diff
@@ -2880,12 +2907,14 @@ L1053               ldu       <D.Proc             get current process pointer
                     cmpb      #S$Intrpt           is it a keyboard interrupt/
```
boisy, line 2907: Probably want "/" to be ?

boisy (changes requested):
Great work here @strickyak and thank you for doing this. One minor nit...

level2/modules/rbf.asm
strickyak: Done

boisy (approved):

PR #42, opened 2024-01-05
krn.asm patch for 6309

ejaquay:
The first commit corrects conditional assemply overlap error in level2 krn.asm that was preventing 6309 boot.  A second commit adds comments to clarify the scope of  conditional assembly in the system block initilization code.

NOTE: Generated F256 code is believed to be correct but this has not been tested.

Hold up - I am reverting the second commit and am trying to get asmprettyprint working instead.

Latest commit removes duplicate section of source per boisy request.

### DIFF
--- a/level2/modules/kernel/krn.asm
+++ b/level2/modules/kernel/krn.asm
@@ -162,6 +162,7 @@ forever@            bra       forever@            simply branch forever
 entry               equ       *
 * Initialize the system block (the lowest 8KB of memory).
 *>>>>>>>>>> F256 PORT
+====if f256
                     ifne      f256
 * F256-specific initialization to get the F256 to a sane state.
                     orcc      #IntMasks           mask interrupts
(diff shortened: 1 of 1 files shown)


level2/modules/kernel/krn.asm
```diff
@@ -203,28 +207,39 @@ l@                  std       ,x++                store the vector value in the
                     bra       l@                  and do again
 done@               ldx       #$0000              start clearing at this address
                     ldy       #$2000              for this many bytes
+                    clra                          set D
```
boisy, line 210: Why are you duplicating code? Lines 209-216 are the same as 232-238. There must be a better way to arrange the conditionals so that code isn't duplicated.

boisy (changes requested):

level2/modules/kernel/krn.asm
ejaquay: Only in source, not the binaries but I can try something else.  Will require adding an "ifeq H6309" instead of what I did.  Meanwhile I am fighting with asmprettyprint.  It seems to be moving all the comments around when they used to be lined up.

ejaquay: Per your request I used IFEQ H6309 to fix the problem instead of my previous method.  This gets rid of duplicated source.

boisy (approved):

boisy:
Thanks for your efforts here. I approved your changes and will merge them in.

PR #43, opened 2024-01-06
Fixed minor build issue.

boisy:
Need to put a blank line between local `l@` label so assembler would work.


ejaquay:
I see no problem with adding this line, it was caused by my lack of understanding about how temporary labels were scoped.

ejaquay (approved):
Thanks. I did not catch it because I did not test F256 build.  Next time I will try that too.

PR #44, opened 2024-03-19
New booter for emulator vhd files

ejaquay:
New booter boot_emu.asm uses boot_common to allow for non-contigous OS9Boot files.  Compare 3rdparty boot_vhd.asm

Note:  This code is idented to the current repo standard.  When a reduced indent spacing standard is implemented this code should be re-indented to suit.


boisy (approved):
Looks good. Thanks for doing this work.

ejaquay:
Added emulator virtual hard drive booter but changes to makefiles would be required to create a boottrack that uses it if automating that process should be desired.

PR #50, opened 2024-07-19
Foenix F256 Port additions

boisy:
There are a lot of changes here impacting mostly the Foenix F256 port, but also other ports.


3rdparty/packages/basic09/basic09.asm
```diff
@@ -478,8 +478,11 @@ L0022               fdb       $1607               Edition #22 ($16)
 
 * Intro screen
 
+                    ifne      F256
+L0024               fcb       $0A
+				else
 L0024               fcb       $0C
```
strickyak, line 484: what does Form Feed do on F256?
and if it clears the screen, why do that?  
( maybe there was something on the screen I wanted to copy into the program. 
Basic should only do its job, and not erase my screen. )

boisy: We've deployed a new banner on the Basic09 version of the F256. The linefeed just puts an extra line after the banner is displayed. We conditionalized out the $0C to clear the screen, which is what the original Basic09 does.

strickyak: SG.  Just curious, what else changes on the F256 version of Basic09?
Does it use the Math Coprocessor feature?

strickyak (approved):
LG

3rdparty/packages/basic09/basic09.asm
boisy: Note, this branch is going to be abandoned. There are too many changes across too many places.

PR #54, opened 2024-08-24
OS-9 in ROM booting fix

boisy:
I added more robust code to `boot_rom.asm` for OS-9 Level 2 on the CoCo 3 that fixed the boot issue.

### DIFF
--- a/level2/coco3/modules/boot_rom.asm
+++ b/level2/coco3/modules/boot_rom.asm
@@ -44,13 +44,22 @@ start               equ       *
                     bcs       Uhoh
 
                     stu       3,s                 save pointer in X on stack
+                    deca                          reduce the number of pages we're returning
                     std       1,s                 save size in D on stack
 
-* TRICK!  Map block 4 into $4000, copy our special ROM copy code
-*         there, then jmp to it!
+* Call F$SRtMem to return the last page of RAM.
+                    leau      d,u
+                    ldd       #1
+                    os9       F$SRtMem
+                    
+RelAddy             equ       $4800
+
+TBLOCK              equ       5
+* TRICK! Map block TBLOCK into $4000 of the kernel address space temporarily, then 
+* copy our special ROM copy code there and jump to it.
                     lda       $FFA2
                     pshs      a
-                    lda       #$04
+                    lda       #TBLOCK
                     sta       $FFA2
                     ifne      H6309
                     ldw       #RelCodeL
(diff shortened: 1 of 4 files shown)


strickyak (changes requested):
level2/coco3/bootroms/makefile
does not end with a \n character.
Please fix, for unix portability.

level2/coco3/modules/boot_rom.asm
```diff
@@ -80,13 +89,17 @@ ExitOK              andcc     #^Carry             clear carry
...
 
-* this code executes at $4800
+* this code executes at RelAddy
 RelCode             equ       *
                     lda       #$4E                CC3 mode, MMU, 32K ROM
                     sta       $FF90
                     sta       $FFDE               ROM/RAM mode
 
 * Map ROM Blocks in
+                    ldd       $FFA6
```
strickyak, line 99: I thought that, in general, reading from $FFAx was not guaranteed to work
(and so the OS always keeps a shadow of these values in RAM somewhere). 

Not requiring this hardware ability (to read at least the low 6 bits from the MMU)
would allow simpler hardware designs for 6x09 SBCs to be coco-MMU compatible.

level2/coco3/modules/boot_rom.asm
```diff
@@ -80,13 +89,17 @@ ExitOK              andcc     #^Carry             clear carry
...
+                    pshs      d
+                    ldd       $FFA4
+                    pshs      d
                     ldd       #$3C3D
```
strickyak, line 103: I see what you're doing here, but I suspect that many readers will not.

And since this is the kind of code that other 6x09 SBC builders
will need to muck with, some comments would help.

level2/coco3_6309/bootroms/makefile
```diff
@@ -39,3 +43,6 @@ kernel_rom: $(KERNEL_ROM) $(DEPENDS)
 clean:
 	$(RM) $(ALLROMS) $(KERNELS)
 
+# Load Ostrich Emulator
+loadromemu: nos96309l2.rom
+	ostrich2.py --device $(ROMEMUDEV) write --address 0x8000 $^
```
strickyak, line 48: There is no file named ostrich2.py

strickyak (commented):
Where is a description of the bug that is fixed?

strickyak (approved):
Approval, if you decide it's not worth changing.

level2/coco3/modules/boot_rom.asm
boisy: I am almost certain that reads from $FFAx work fine. The OS doesn't keep shadows of $FFAx, although some registers in the $FF9x area are write only.

level2/coco3/modules/boot_rom.asm
boisy: Fair point. I'll add some comments.

PR #55, opened 2024-09-12
The scdwv driver now supports the SS.BlkWr setstat call

boisy:
This change allows a program to open up a /n device and use this SetStat to blast bytes to the dwio subroutine module.

### DIFF
--- a/level1/modules/scdwv.asm
+++ b/level1/modules/scdwv.asm
@@ -247,7 +247,7 @@ Write               equ       *
                     pshs      a
                     leax      ,s
                     ldy       #$0002              ; 3 bytes to send.. ugh.  need WRITEM (data mode)
-Blast               ifgt      Level-1
+                    ifgt      Level-1
                     ldu       <D.DWSubAddr
                     else
                     ldu       >D.DWSubAddr
@@ -488,8 +488,8 @@ SendStat
 SetStat
                     cmpa      #SS.Open
                     bne       isitblkwr
-                    lbsr       open
-                    lbcs       ssbye
+                    lbsr      open
+                    lbcs      ssbye
                     ldd       #SS.Open*256+OP_SERSETSTAT
                     bra       SendStat
 isitblkwr           cmpa      #SS.BlkWr
--- a/level1/modules/scdwv.asm
+++ b/level1/modules/scdwv.asm
@@ -247,7 +247,7 @@ Write               equ       *
                     pshs      a
                     leax      ,s
                     ldy       #$0002              ; 3 bytes to send.. ugh.  need WRITEM (data mode)
-                    ifgt      Level-1
+Blast               ifgt      Level-1
                     ldu       <D.DWSubAddr
                     else
                     ldu       >D.DWSubAddr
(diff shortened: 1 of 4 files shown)


strickyak (approved):
comments, but no show stoppers.   Approved.

level1/modules/scdwv.asm
Deek, line 1: Looks like it's missing the Driver ID getstat that allows user space to know whether block I/O is actually possible. From my defs file for s16550:
```
* Additional SIO GetStat/SetStats
SS.TxBuf	equ	$C1	Get statistics about the Tx buffer (org driver called it SS.WrtBf)
SS.TxFlR	equ	$C2	Reset (clear) TX software flow control
SS.TxClr	equ	$C3	Clear transmit buffer contents. (SS.TxFlh)
SS.RdBlk	equ	$D0	Block Read (name changed for compatibility) (SS.BlkRd
SS.WrBlk	equ	$D1	Block Write (same)
SS.DvrID	equ	$D2	Serial driver interrogation

* These are from SACIA et al.
DCDStBit	equ	%00100000	DCD status bit for SS.CDSta/ComSt calls
DSRStBit	equ	%01000000	DSR status bit for SS.CDSta/ComSt calls```

s16550 docs: ```
SS.WrtBf  returns  interesting stats about the transmit  buffer.
   $C1    Reg  X  contains the buffer size, Reg Y  contains  the
          current fullness (number waiting to be sent), and Reg B
          indicates  the status of flow control. If B=0, no  flow
          control.   If B<>0 then flow control has been requested
          and the transmitter is off.

SS.BlkRd  Acts  like  SS.Ready  if there  is  no  data  waiting.
   $D0    (E$NotRdy  returned) Acts like a SS.Read  (kinda)  if
          data  is  present. It will return the number  of  bytes
          requested,  or  the  number available,  which  ever  is
          smaller.   It will *never* go into suspend state.  This
          call  copies data from the driver's buffer directly  to
          your buffer, bypassing SCF.  Note that the use of BlkRd
          eliminates the need for a seperate SS.Ready.
                Entry     A = path
                          X = pointer to buffer
                          Y = maximum nuber of bytes to transfer
                Exit      X = pointer to buffer
                          Y = number of bytes moved (Y<>0)
                Errors    B = E$NotRdy or E$Read

SS.DvrID  returns a few predined constants. So far:
   $D2
          reg X
           0 = supports SS.hangup ala SAcia and s16550
           1 = supports SS.Break ala sacia
           2 = supports signal on DCD change
           3-7 undefined
           8 = has ss.blkrd
           9 = has ss.blkwr
           a = has tx-flush
           b-f undefined
                reg b contains bitmapped driver/chip type
                  0 = 6551
                  1 = 6552
                  2 = 1655x

                reg Y contains the driver id number
                reg A the lower 4 bits contain the max baud rate number.
                      Not  the  number  of baud rates,  but  the  highest
                      allowable number (currently $b)```

It should be noted that the SACIA block I/O stats don't have the same numbers as the RBF block read/write stats, that's why I renamed them to SS.(Rd|Wr)Blk. There is software that is already aware of the $D0, $D1, and $D2, and they automatically enable block I/O if getstat $D1 returns the appropriate bits in X.

**Comment by MatthewMassie** [comment]
Looks good to me.

File: level1/modules/scdwv.asm
Comment (Deek) [reply level1/modules/scdwv.asm:1]: I should also note that either the above documentation or the code for getstat $D2 is wrong, because the actual 16550 driver returns $0077 instead of $0707. I haven't checked what SACIA 6551/6552 return from $D2 so I don't know which is the case.

Comment (boisy) [reply level1/modules/scdwv.asm:1]: I don't see these stat calls in the current repo.

File: level1/modules/scdwv.asm [hunk 1]
```diff
@@ -487,11 +487,67 @@ SendStat
 *
```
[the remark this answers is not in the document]
boisy: Yeah, we really need to document individual SS calls.

PR #62, opened 2024-12-20
VRN docs for the docs/ dir, from an ar2 file.

strickyak:
See my notice at the top of vrn.txt for where the
two included documents were found.

Strick had never seen these files before, because `find` and `grep` and search engines would not find them in the ar2 file.


Deek:
Looks good to me.

Deek (approved):
Looks good.

PR #63, opened 2025-01-14
Add F256 specific banner to basic09

boisy:


Deek (approved):
I don't see anything particularly wrong with this. It might be nice to have one with an extremely minimal banner too, especially if dumb terminals are in use, but fancy looking stuff is fine too.

MatthewMassie (commented):
Changes look good. I do have early Ascii works for NitrOS-9 banner in 80 column that might be nice for terminal based use.

PR #66, opened 2025-01-14
Promote bawk and more to a standard level 1 command

boisy:
Improve comments on commands in level 1


Deek:
This should be more than one commit, it's not like all of this stuff was done simultaneously.

You recommented and "source-ified" some source, removed help from others, changed the stack reservation of some utilities, _and_ moved them into the built images.

PR #68, opened 2025-01-14
Add comments to commands for clarity

boisy:


level1/cmds/shell_21.asm
```diff
@@ -28,52 +26,45 @@
 tylg                set       Prgrm+Objct
 atrv                set       ReEnt+rev
 rev                 set       $00
-edition             set       22
+edition             set       23
```
Deek, line 29: I can't tell if this is a pure comment update or if it has actual code changes that might break something and therefore needs an edition revision. Which is it?

level1/cmds/echo.asm
```diff
@@ -26,18 +24,18 @@ edition             set       5
                     mod       eom,name,tylg,atrv,start,size
 
                     org       0
-u0000               rmb       450
+                    rmb       200       stack
```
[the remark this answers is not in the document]
boisy: 450 bytes of stack size is excessive for this small program. Typically stack sizes are around 200 bytes. There are no static variables used, so reducing the stack saves on the amount of memory requested at fork time.

level1/cmds/list.asm
```diff
@@ -25,10 +23,12 @@ edition             set       5
 
                     mod       eom,name,tylg,atrv,start,size
 
+maxreadlen          equ       200
                     org       0
 filepath            rmb       1
 parmptr             rmb       2
-readbuff            rmb       650
+readbuff            rmb       maxreadlen+1
+stack               rmb       200
```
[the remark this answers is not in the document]
boisy: Correct. This is to optimize the memory request size and delineate the stack requirements from the read buffer in an obvious way.

level1/cmds/shell_21.asm
```diff
@@ -99,93 +91,113 @@ IcptRtn             stb       <kbdsignl
...
-                    beq       L00AA
-                    bsr       WriteLin
-L00AA               lbsr      L017B
-                    bcc       L0069
-                    tstb
-                    bne       L0087
-                    bra       L0069
+                    bne       ExitShell        exit shell if so
+noparms@            lds       ,s++             get Y off stack pushed earlier (end of parameter area)
+ProcLine            tst       <suppressintro   do we suppress showing the intro?
```
boisy, line 123: @Deek a subtle optimization... I now test the "supressintro" flag before setting X two lines down. The other code set X first and then did the test. It may not justify incrementing the edition byte. Let me know what you think.

level1/cmds/shell_21.asm
boisy: See my comment below.

Deek: I would say if it's a pure (micro-)optimization that shouldn't be able to break anything, there's probably no need for a new edition.

But that's also why it's not great to make code and formatting (like comments) changes at the same time, because the tiny code update/fix/etc. that actually alters the binary gets hidden by the big line count changes that don't.

boisy: I reverted the edition back to 22.

PR #70, opened 2025-01-14
Add pick.as to level1/cmds

boisy:
- pick was also added as a target to `f256/level1/cmds/makefile`. Since it's only used for the F256 port, it only resides in that makefile, but is also potentially useful for other ports, which is why the assembly source is in the overall `level1/cmds` folder.

### DIFF
--- a/level1/cmds/pick.as
+++ b/level1/cmds/pick.as
@@ -1,0 +1,187 @@
+********************************************************************
+* pick - Scripting Utility
+*
+* Edt/Rev  YYYY/MM/DD  Modified by
+* Comment
+* ------------------------------------------------------------------
+*   1      2024/06/28  Boisy Pitre
+* Created.
+
+scrdelim        set     '|
+maxlinelen		set		80
+
+                    section   bss
+keypressed          rmb       1
+isflash             rmb       1
+prompt              rmb       2
+scriptpathlist      rmb       2
+linebuff		    rmb		  maxlinelen+1
+                    rmb       100
+                    endsect
+
+KeySig              equ       $84
+
+                    section   code
+                    
+IcptRtn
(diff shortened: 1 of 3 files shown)


level1/cmds/pick.as
```diff
@@ -0,0 +1,189 @@
...
+* Pick Scripts
+*
+* Pick scripts are text files that are parsed to perform some command based on a key.
+* A script is made up of one or more delimited lines of this form:
+*
+* K|M|C
+*
+* Where:
+*    K = the key used to invoke the command
+*    D = a description string which describes the command
```
strickyak, line 108: M (above) or D (here)?

level1/cmds/pick.as
```diff
@@ -0,0 +1,189 @@
...
+                    sta       keypressed,u
+                    rti
+
+DefPrompt           fcc       ":"
+                    fcb       $00
+                    
+**********************************************************
+* Entry Point
+**********************************************************
+__start             cmpd      #$0001              any parameters?
```
strickyak, line 39: Document how to invoke the command -- what is -p?  where is script name?

strickyak (changes requested):
also, isn't there a HELP directory, you could document it in?

level1/cmds/pick.as
boisy: Fixed in the latest push.

boisy:
> also, isn't there a HELP directory, you could document it in?

I also added `pick.hp`.

level1/cmds/pick.as
boisy: I made the comments more robust and added `pick.hp`

strickyak (approved):

PR #71, opened 2025-01-14
Minor kernel enhancements

boisy:
- Fix comment in ffork.asm regarding pages vs bytes
- Level 1 kernel now places the `DUMMY` pointer in the  `D.NMI` vector .This addresses an issue that caused a crash on the F256. It's a good idea for all ports to do this.


strickyak (approved):
LGTM

PR #73, opened 2025-01-14
Add Foenix sound file library

boisy:
This change affects only the Foenix port. A new library has been added for interacting with sound devices.

### DIFF
--- a/level1/f256/libs/fnxsnd/psg.as
+++ b/level1/f256/libs/fnxsnd/psg.as
@@ -1,0 +1,100 @@
+                    section   .bss
+nega                rmb       1
+                    endsect
+
+                    section   .text
+
+MAPSLOT             equ       MMU_SLOT_1
+MAPADDR             equ       (MAPSLOT-MMU_SLOT_0)*$2000
+
+PSG.Base            equ        PSGL.Base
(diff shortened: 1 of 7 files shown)


level1/f256/libs/fnxsnd/psg.as
```diff
@@ -0,0 +1,92 @@
...
+* Entry: D = Value to send to CODEC.
+*        X = Base address of CODEC.
+SendToCODEC         
+                    std       CODECCmdLo,x
+                    lda       #$01
+                    sta       CODECCtrl,x
+l@                  cmpa      CODECCtrl,x
+                    beq       l@
+                    rts
+
```
strickyak, line 80: is PSG_WRITE a public entry, like PSG_INIT?  With an argument at 2,s?  Document?

strickyak (changes requested):
PSG_WRITE might need documentation

level1/f256/libs/fnxsnd/psg.as
boisy: It is public. The byte is passed in `A`.

strickyak (approved):

PR #75, opened 2025-01-17
Fix build of libfnxsnd.a and paths that reference it.

boisy:
I moved libfnxsnd.a from the root folder's lib folder into level1/f256/libs since it's an F256-specific library.


jfed6000 (approved):
Reviewed.

PR #81, opened 2025-01-18
HDBDOS flag conditional guard

boisy:
Place conditional guard around the HDBDOS flag which is not needed for platforms like the F256. This keeps the size down by eliminating unused/irrelevant code for those platforms.


MatthewMassie (approved):
Reviewed.

strickyak (approved):
LGTM

Deek (approved):

PR #84, opened 2025-01-20
Adding RealCoCoFPGA

Nitrobotics:
The RealCoCoFPGA is the parent core that the Matchbox CoCo, MiSTer CoCo, SiDi CoCo, MiST CoCo, is based on.  At this time the only thing different about a RealCoCoFPGA is it's custom hybrid SDC/FDC circuit, so it requires it's own OS-9 RBF driver in order to boot, of course.  My port already works and builds nice VHD images, but I'm adding to the repo in parcel to better understand the PR and Merge flow.


boisy (commented):
Roger, the only thing I see here is the addition of a `makefile`. Are there any additional sources or other files that didn't make it into this commit that should have? Just want to confirm. I also don't see a `level2/realcocofpga/cmds` and other directories that I would associate with a port.

Nitrobotics:
I didn't know I could commit it all at once, so I was picking off the 33 or so changes after I dropped in my port's folder.

boisy (approved):
There are some efficiencies that could be gained by referencing the individual `w?.asm` files from `level2/coco3/modules` instead of adding them here, but those can be addressed in another PR.

PR #85, opened 2025-01-20
Wrapping up initial addition for the realcocofpga

Nitrobotics:


level2/coco3/modules/boot_realcocofpga_sdc.asm
boisy, line 1: This file should reside in `level2/realcocofpga/modules`

level2/coco3/modules/makefile
boisy, line 1: I think the right course of action here is to have all `realcocofpga` related files and makefiles isolated into the `level2/realcocofgpa` folder.

PR #87, opened 2025-01-21
Realcocofpga nonintrusive additions

Nitrobotics:
Self-contained approach to my port.  Only one file outside of my dir is modified: level2/makefile


Nitrobotics:
I'll deal with the term, w, w1...w15 duplication soon so long as the existing repo's copies have the conditionals to accommodate more generic symbols rather than COCO3FPGA and ALTCOLORS, etc. which seem to be specific to one port.

boisy (approved):

PR #89, opened 2025-01-25
level1/d64: update makefile

six809:
vpath and flag variables adapted from level1/coco1 changes.

Seems to fix the build for d64 and related ports.


strickyak (approved):
LGTM, d64!

boisy (approved):

PR #96, opened 2025-05-03
Add Microware-specific comments to shell_21.asm

boisy:
Comments from Microware shell.asm source code are at https://www.roug.org/retrocomputing/os/os9


Deek (changes requested):
Committed with CR line endings, the diff is not useful

PR #97, opened 2025-05-04
Add original Microware comments to shell_21.asm

boisy:
- Also code now uses "one space" rule (which is where all code is going soon).
- Verified changes still assemble to original shell_21 size and CRC.

### DIFF
--- a/level1/cmds/shell_21.asm
+++ b/level1/cmds/shell_21.asm
@@ -6,23 +6,23 @@
 * Edt/Rev  YYYY/MM/DD  Modified by
 * Comment
 * ------------------------------------------------------------------
-*  17      1983/01/20  KKK
+*  17      1983/01/20  Kim Kempf
 * Fix RESET to properly DUP paths on errors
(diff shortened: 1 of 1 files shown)


Deek (approved):
Could still use some dedisassemblification (heh), e.g. there's no point to `cmpd #$0000` or `leax -$05,x` other than to say "a program spit this out", but it's more readable.

level1/cmds/shell_21.asm
```diff
@@ -6,868 +6,1149 @@
 * Edt/Rev  YYYY/MM/DD  Modified by
```
strickyak, line 6: Personally, I would appreciate it if we got rid of the RCS $Id$ marks.
Those mess with my workflow occasionally. 
(Yes, I use RCS occasionally for temporary snapshots.)

strickyak:
> * code now uses "one space" rule

I missed the "one space" discussion.
Is there a record of if?

PR #102, opened 2025-05-07
Makefile adjustments

boisy:
- Add removal of *.list and *.map to `clean` targets in makefiles
- Address build location of libfnxsnd.a in f256 makefile

### DIFF
--- a/level1/atari/cmds/makefile
+++ b/level1/atari/cmds/makefile
@@ -38,7 +38,7 @@ pxd:	pd.asm
 	$(AS) $(AFLAGS) $< $(ASOUT)$@ -DPXD=1
 	
 clean:
-	$(RM) $(ALLOBJS)
+	$(RM) $(ALLOBJS) *.list *.map
 
 identify:
 	$(IDENT_SHORT) $(ALLOBJS)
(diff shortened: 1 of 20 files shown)


Deek (approved):
Seems fine.

strickyak:
Be careful, there was either a .list or .map file somewhere under third party software that was checked in.
we should rename it so it doesn't match the pattern.

here it is
     3rdparty/packages/uucpbb/sample_config/bin.list

strickyak:
BTW, it's coco-shelf that adds a shell script wrapping lwasm that adds options to create so many .list and .map files.
Because I have tracing tools that collect and use them. 
I'll soon be adding this automatic collection to coco-shelf, for the TFR/9.

( Other projects are moving this direction, too -- you can give MAME one listing file,
and one of the vloggers is wanting some debug info to pass through lwlink.
But mine is the most automatic, in that it finds the correct checksumed listing files for OS9 modules on its own. )

boisy:
> Be careful, there was either a .list or .map file somewhere under third party software that was checked in. we should rename it so it doesn't match the pattern.
> 
> here it is 3rdparty/packages/uucpbb/sample_config/bin.list

The UUCPbb makefile doesn't remove list or map files, so we're ok.

PR #103, opened 2025-05-07
Update clock2_f256.asm

boisy:
HOUR and HOUR_ALARM parameters of RTC need to consider that bit 7 denotes AM/PM.


Nitrobotics (approved):
Looks good

PR #119, opened 2025-05-14
Change FEU font to banner so that L2 doesn't need to include it.

boisy:


MatthewMassie:
Update looks good. I notice the pallete is different when I boot. I wonder if we will need to update the palette file as well.

boisy:
> Update looks good. I notice the pallete is different when I boot. I wonder if we will need to update the palette file as well.

Yeah, that's possible. I only see one palette file though in the repo, in `level1/f256/modules/palette.asm`

MatthewMassie (approved):
Reviewed.

PR #130, opened 2025-05-17
Update rbmem.asm

Nitrobotics:
Enable the Flash write efficiency.  This may need to become a dmode option for long-term testing.  In my tests it never fails, and provides much faster writes after the Flash chip has been erased or soft-formatted to $FF sectors.


jfed6000 (approved):
Approved.

PR #132, opened 2025-05-18
F256/palette fix

MatthewMassie:
This fixes the 0 palette color issue when booting drive images from FEU. Fixed Basic09 Paint example now that CMDS is cleaned up. Reflashed FEU and tested.


jfed6000 (approved):
Approved

PR #133, opened 2025-05-18
Remove pipe-related modules to keep the FEU bootfile under the necess…

boisy:
…ary size.

- Since FEU is really just a minimal OS-9 Level 1 environment for booting, running the debugger, and basic file manipulation, it doesn't need all the features of a full OS-9 Level 1 system, which can still be made on an SD card.


jfed6000 (approved):
Approved.

PR #141, opened 2025-05-23
F256/wizfi

Nitrobotics:
Reduced driver size.  Compatible with any proposed register base.
xmode /wz hpa=4040  <- current MMU-based scheme
xmode /wz hpa=ff20   <- proposed base being tested


jfed6000 (approved):
Approved

PR #150, opened 2025-05-26
Remove $Id$ and "Disassembled" lines across the source base.

boisy:
This commit removes these longstanding comment lines in sources:

```
* $Id$
*
```

and:

```
* Disassembled ...... by RML (or whomever)

```


level1/cmds/dmode.asm
```diff
@@ -44,13 +42,9 @@
 * May be a back-port from Carl Kreider's dmode for OSK
 *
 * 2 / 4    2004/02/28  Rodney V. Hamilton
-* Disassembled new dmode, added comments and KD/RAK mods.
```
Deek, line 47: This is a real comment.

Deek (changes requested):
I have no problem removing $Id$ lines, but Disassembled lines include useful information.

They don't just say when it was done either, some of them include the names of the people who did the actual work of disassembling and documenting the code. That information should not be deleted, and if it's already been removed before it should be restored -- I notice that an earlier commit stripped a notice from Alan Dekok.

level1/cmds/dmode.asm
boisy: Good point. I forced push a new commit that only removes $Id$.

PR #153, opened 2025-05-29
start docs/history.txt with prose from LCB

strickyak:
What do folks think of starting a text file to record the history of NitrOS9?
To save tidbits of history that come out on the Discord or other social media.
We can save scraps here, hoping someone will incorporate them into a better work. 

If we decide to keep it, consider this free to be added to, like a wiki page.


boisy:
> What do folks think of starting a text file to record the history of NitrOS9? To save tidbits of history that come out on the Discord or other social media. We can save scraps here, hoping someone will incorporate them into a better work.
> 
> If we decide to keep it, consider this free to be added to, like a wiki page.

I think it's a great idea.

boisy (approved):

PR #160, opened 2025-06-07
level1/d64 - Re-enable /t1 using the ACIA51 driver.

andrewcsteele:
This uses the unchanged source for ACIA51 from:
	https://www.roug.org/retrocomputing/os/os9/l1sources/acia51

and adds the device descriptor to all bootfiles, the driver added to CMDS
 where it may be loaded only when required.


Deek:
Is there any benefit to adding a device descriptor to the boot file if the driver is missing? Without the driver it can't be iniz'ed, and it's just as easy to append the descriptor to the end of the driver file.
```
load acia51
iniz t1```
will still work, and you don't have the errors attempting to initialize a device without a driver.

andrewcsteele:
The Dragon Data boot disk has t1 present at boot and acia51 in cmds, that was my goal.

I tried 
merge acia51 t1_acia51.dt > foo
load foo
and t1 shows in mdir and xmode /t1 and tsmon /t1 & both appear to work without the iniz, but more I can see no other drivers/descriptors treated this way?

Deek:
> The Dragon Data boot disk has t1 present at boot and acia51 in cmds, that was my goal.

OK, if the original boot disk was dumb too I suppose it kinda makes sense to carry forward the silly decision for "compatibility"

PR #161, opened 2025-06-11
Add TFR9 as a choice within the coco1 port.

strickyak:
This could be its own new PORT=tfr9, and probably
that will be the right idea later on.

But for now, rather than duplicate a dozen or so files, I can "inherit" from coco1 with a much smaller change, which essentially does two things:
    -- It chooses EMUDSK as the /DD.
    -- It chooses SC6850 at $FF06 as the /TERM.

I'm currently revamping the makefiles from the hasty, kludged /v2/ to a supported, repeatable /v3/ firmware release inside https://github.com/strickyak/tfr9/ and it will depend on a configuration like this being in nitros9.

Thanks!


boisy (changes requested):
Henry, I am against putting the TFR9 into the CoCo 1 port. Regardless of similarity, it really should be its own port. Can you redo this so it resides in its own port folder/

strickyak:
OKAY, I'll go with Boisy's suggestion, and make the `tfr9` its own PORT.   But at the moment, the /v3/ build depends on this branch, so I'll keep the branch frozen here, until the new way is done.

PR #162, opened 2025-06-12
Add tfr9 as a choice within coco3 (and coco1 also)

strickyak:
This depends on the other PR that adds a Level 1 named tfr9 inside coco1.

This adds a commit to add a Level 2 similarly, inside coco3.


strickyak:
OKAY, I'll go with Boisy's suggestion, and make the `tfr9` its own PORT.   
[ See https://github.com/nitros9project/nitros9/pull/161 ]
But at the moment, the /v3/ build depends on this branch, so I'll keep the branch frozen here, until the new way is done.

strickyak:
Deleting this request.  I have an experimental new way, using a Tcl script to create one Makefile in a build directory. https://github.com/strickyak/tfr9/blob/main/v3/rmake.tcl

PR #163, opened 2025-06-13
Add support for new text lut addr for RC0017_002+

jfed6000:
Add support for new location of Text LUTs in RC0017_002

### DIFF
--- a/level1/f256/modules/vtio.asm
+++ b/level1/f256/modules/vtio.asm
@@ -286,10 +286,14 @@ l@                  tfr       d,x                 transfer it to X
                     bcs       installfont         branch if the link failed
                     pshs      y                   save Y
                     tfr       y,x                 transfer it to X
-                    ldy       #TEXT_LUT_FG        load Y with the LUT foreground
+		    lda	      #TEXT_LUT_BLK	  load text LUT block
+		    sta	      MAPSLOT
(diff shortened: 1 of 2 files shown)


MatthewMassie:
Reviewed.

level1/f256/modules/vtio.asm
```diff
@@ -346,7 +350,7 @@ l@                  ldd       ,x++                get two bytes from the source
 * Clear memory at MAPADDR with the contents of D.
 clr                 ldx       #MAPADDR
 l@                  std       ,x++
-                    cmpx      #MAPADDR+80*61
```
boisy, line 349: Why did this change from 61 to 60? Was it a bug initially?

boisy (changes requested):

level1/f256/modules/vtio.asm
jfed6000: Sort of, it writes 1 extra time when you don't need it.  It is zero indexed.  So if it writes to 0+4800, that is actually just past the end of the screen which is 0 + 4799.  It was causing problems when the CLUT was in $C2, but the CLUT is now in $C1.  Probably doesn't hurt anything, but it clears the screen just fine.

boisy (approved):

PR #166, opened 2025-06-15
Add s1 descriptor to feu makefile

boisy:


boisy:
This change simply adds the `s1` descriptor to the OS-9 Level 1 boot file in FEU.

jfed6000 (approved):
Approved

PR #171, opened 2025-06-16
Remove duplicate target in makefile

boisy:


boisy:
> upload_f256jr: bootfile_dw_f256jr upload bootfile_f256kr 8000 <---- typo? should be bootfile_f256jr ?

Yep, actually that's supposed to be `upload_f256k` and `bootfile_f256k`. I shouldn't have deleted that target since it's for the K. I added it back in and fixed it.

PR #177, opened 2025-06-19
Update mergeln.asm

Nitrobotics:
Mergeln exits gracefully if the SCF driver returns a Not Ready status after 3 seconds.


jfed6000 (approved):
Approved

strickyak:
Why?   Does this make it nondeterministic? 
Why should the result of a copy command depend on how fast your computer is or how busy it is?

PR #180, opened 2025-06-28
F256/wizfitool

MatthewMassie:
This adds wizfitool.b09 to Level2 cmds folder and required files for background. This allows basic configuration of the WizFi360, list APs, Ping devices, Connect to AP, Disconnect from AP, Access Remote Login and DW, DHCP/Static IP, Wizfi firmware updater.


Nitrobotics:
The Remote Login feature is failing for me.  For the mouse option I can see the X,Y coordinates updating, but clicking a menu item doesn't do anything.

boisy (changes requested):
I believe we should put `wizfitool.b09` in `level1/f256/cmds` since it's an F256-specific tool.

MatthewMassie:
> The Remote Login feature is failing for me. For the mouse option I can see the X,Y coordinates updating, but clicking a menu item doesn't do anything.
I have not had any mouse issues but try making sure the tip of mouse is directly pointing to option you want to select and left click.

boisy (approved):

PR #183, opened 2025-07-08
Add zero-key check to ps/2 keyboard so that key returning zero from s…

jfed6000:
add zero-key check to ps/2 keyboard so that key returning zero from scanmap doesn't send abort to process


MatthewMassie:
Reviewed.

PR #192, opened 2025-07-29
added buffer for KySns modifier keys

jfed6000:
This changes f256vtio.d to add a variable for the KySns buffer and supporting code in keydrv_ps2 and vtio.  This lets the program get both the key and the modifier that was pressed at the time from the buffer.


MatthewMassie (approved):
Reviewed. Approved.

PR #193, opened 2025-08-16
Fix build errors

tlindner:
I ran into problems doing a full build. This fixes the issues I ran into.


strickyak (approved):
LG

tlindner (commented):
Looks good.

PR #196, opened 2025-08-24
minor lbra fix to modem, vtio fix for reverse video

jfed6000:
Added lbra instruction to line 198 of modem.asm to fix compile issues.

vtio reverse video needed to reset to the default handler after setting the reverse video.  Prior version would try to read a second parameter if two reverse video calls were made in a row.


MatthewMassie (approved):
Reviewed  and Approved.

tlindner:
The `modem.asm` fix was a dupe of PR #193.

PR #269, opened 2026-01-18
Adjust wildbits libraries for level 1 and level2.

DrPitre:
We are no longer just creating `libwildbits.a.` We're now building `libwildbitsl1.a` and `libwildbitsl2.a` so that Level 1 and Level 2 version of the Wildbits port can link to proper symbols.


jfed6000 (approved):
Approved.  New library structure.

PR #270, opened 2026-01-19
Modify version so that it shows date for Wildbits port

DrPitre:
The Wildbits OS-9 team wants to move to a date-based version scheme (e.g. 26.01.18). This affects the `init` module which carries the version numbers both as embedded binary values and part of the version string.

We don't want to affect other ports, so I made the values conditional in `rules.mak`.

- Wildbits port now reports version as YY.MM.DD (e.g. 26.01.18)
- Other ports remain at 3.3.0
- As a result of modifying `init.asm`, its embedded version string now shows two digits for all ports (e.g. `03.03.00` instead of `3.3.0`)

### DIFF
--- a/3rdparty/packages/pacos9/makefile
+++ b/3rdparty/packages/pacos9/makefile
@@ -7,6 +7,13 @@ LFLAGS += -L$(NITROS9DIR)/lib -lnos96809l2
 
 DSK = PacOS9.dsk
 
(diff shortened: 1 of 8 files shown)


strickyak (approved):
Approve commit 6099f5ba, with `ifgt      NOS9VER-25`.
   strick

PR #280, opened 2026-03-01
Update build to cache lwtools

callsop:
This will update the build check to cache lwtools so that its not rebuilt every time there is a build. Retention days is specified in the settings/actions/general under title 'Artifact and log retention'. The default is 90 days, this means if there is no build within 90 days, the cached tools will be re-fetched.

This should almost eliminate the chance of a bad build due to a network issue.

There is instructions for updating to newer lwtools in the lwtools.sh file.

This will also speed up the builds by a few seconds.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -26,6 +26,9 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+
+#
+# Prepare environment with necessary packages.
 # disabled for now, uncomment if need to install package
 #    - name: Setup Requirements
 #      run: |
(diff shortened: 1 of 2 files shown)


callsop:
You can see sample of this here: https://github.com/nitros9project/nitros9/actions/runs/22602465290/job/65487224838 

Skipping the lwtools fetch:
image

callsop:
@boisy Now that you have recipes in I think you can use this.

callsop:
FYI, its not out of date, the affected files remain unchanged/can be merged.

DrPitre (approved):

callsop:
@boisy Thanks. Someone else will need to commit my prs as I don't have that permission. :)

PR #287, opened 2026-03-28
Recipe integration

DrPitre:
This PR integrates what was in the NitrOS-9 recipes repo directly into this repo.


strickyak (commented):
Thanks for checking in Recipes without deleting the Makefiles that
coco-shelf and other things depend upon.   Sorry I haven't had time
to study what's going on here yet, but I appreciate the idea, since I
had created my own recipe-like experiment.  

You remember that I was trying to run with mostly standard modules,
but picking and choosing different combinations of peripherals than were the
standard one true way of booting that platform.   I'm hoping the new scheme
will let me inherit an existing recipe with small changes or recombination
of modules, with a minimal description of the differernces.   Or at least, 
that it is not as heavyweight as producing a new "port" was in the past. 

I should give this a good look ....

strickyak:
The coco/40d recipe depends on file `defs/buildinfo` but does not know to make it.
So recipes don't work on a clean checkout.

A related bug:  in the top level directory of nitros9, `make clean` does not delete `defs/buildinfo`,
so it doesn't leave you a clean checkout.

boisy:
> The coco/40d recipe depends on file `defs/buildinfo` but does not know to make it. So recipes don't work on a clean checkout.
> 
> A related bug: in the top level directory of nitros9, `make clean` does not delete `defs/buildinfo`, so it doesn't leave you a clean checkout.

Thanks. I addressed these in the latest commit.

strickyak (approved):
I like it! 
See my suggestion elsewhere, about NITROS9DIR.
   -- strick

PR #295, opened 2026-04-26
wildbits: fix wrong naming in bootfile

DrPitre:
Summary: fix WildBits linked-command module names so OS-9 headers do not pick up the output path like .mods/sysgo.

Changes:
- link RMA/RLINK-built modules to the bare basename first, then move them into .mods
- apply the same fix to the FEU-specific sysgo rule

Verification:
- rebuilt the WildBits FEU bootfile
- verified with os9 ident -s bootfile that sysgo, wbreset, and bootos9 no longer carry a .mods/ prefix in their module headers


MatthewMassie (approved):
Reviewed. Looks good.

PR #296, opened 2026-04-26
wildbits: add basic09 to disk images

DrPitre:
Summary: add Basic09 commands, sample programs, scripts, and tests to WildBits disk images.

Changes:
- add Basic09 package source/include path to the shared WildBits recipe
- include basic09, runb, inkey, and syscall in CMDS
- copy Basic09 sample .b09 files into a BASIC09 directory on the image
- copy WildBits scripts and tests into SCRIPTS and TESTS directories
- make the Level 2 image depend on sysgo so the packaging step succeeds

Verification:
- dry-run checked WildBits Level 1 and Level 2 recipe builds
- verified the Level 2 jr2 build path after fixing the shared recipe dependencies

### DIFF
--- a/recipes/wildbits/wildbits.mak
+++ b/recipes/wildbits/wildbits.mak
@@ -4,6 +4,8 @@ MODDIR = .mods
 include ../../rules.mak
 -include recipe.mak
 
+vpath %.asm $(3RDPARTY)/packages/basic09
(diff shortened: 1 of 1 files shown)


MatthewMassie (approved):
Approved.

PR #297, opened 2026-04-26
wildbits: copy startup into disk images

DrPitre:
## Summary
- copy the WildBits startup file into Level 1 and Level 2 disk images
- make the shared WildBits disk image targets depend on the startup source
- preserve the existing BASIC09, scripts, and tests packaging flow

## Testing
- make -C recipes/wildbits/l1 PLATFORM=jr -n | rg "startup|OS9Boot"
- make -C recipes/wildbits/l2 PLATFORM=jr2 -n | rg "startup|OS9Boot|sysgo"


MatthewMassie (approved):
Approved.

PR #299, opened 2026-04-28
wildbits: keep cursor color in sync after writes

DrPitre:
## Summary
- update WildBits vtio attribute writes to advance to the next attribute cell after output
- seed the next cell with the active foreground/background color for the cursor
- avoid writing past the end of the largest text buffer

## Testing
- not run


MatthewMassie (approved):
Approved.

PR #300, opened 2026-04-29
wildbits: clarify vtio block mappings

DrPitre:
## Summary
- replace hard-coded WildBits VTIO block values with named symbols
- add a bitmap block symbol and use it in bitmap register access paths
- clean up nearby comments to match the symbolic block usage

## Testing
- not run


MatthewMassie (approved):
Approved.

PR #301, opened 2026-04-29
wildbits: streamline vtio escape dispatch

DrPitre:
## Summary
- simplify the VTIO 1B escape dispatch branches for the palette and DriveWire handlers
- collapse the font set selection handlers to share the final register write path
- keep behavior the same while trimming unnecessary long branches

## Testing
- not run


MatthewMassie (approved):
Approved.

PR #303, opened 2026-04-30
Optimize MMU address calcs

Nitrobotics:
Save ~33 CPU cycles

### DIFF
--- a/level1/wildbits/modules/vtio.asm
+++ b/level1/wildbits/modules/vtio.asm
@@ -1623,7 +1623,7 @@ ex@                 rts
 
 ;;; difference between get and set is just two lines specifying
 ;;; source and destination.  So procedures are combined.
-GSFntChar           lda       #0
+GSFntChar           clra
                     bra       DoFontGetSet
 SSFntChar           lda       #1
 DoFontGetSet        pshs      a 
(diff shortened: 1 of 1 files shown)


level1/wildbits/modules/vtio.asm
```diff
@@ -1843,10 +1843,12 @@ map@                lda       R$Y+1,x             load bitmap@
...
                     ldb       R$Y+1,x             ldb with bitmap#
-                    lda       #$08                multiply by 8 (to start at 0, 8 or 16)
-                    mul                           d should be 0,8,or 16
-                    addd      #MAPADDR
-                    addd      #$1000
+                    clra
+                    lslb                           multiply by 8
+                    lslb
+                    lslb
+                    adda      #MAPADDR/256
```
DrPitre: Could you remove the `clra` on line 1846 and change this line to `lda #MAPADDR/256+$1000/256` and eliminate the next `adda` line for the same effect?

level1/wildbits/modules/vtio.asm
```diff
@@ -2024,14 +2026,16 @@ SSPalet             pshs      cc
...
```
DrPitre: Same comment here as above.

DrPitre (changes requested):

level1/wildbits/modules/vtio.asm
DrPitre: There's also an opportunity to save a byte by changing the `lda #0` on line 1860 to `clra` and there are other places in the code that does this.

DrPitre (approved):
Great optimizations!

level1/wildbits/modules/vtio.asm
Nitrobotics: Done.  I see potential for someone else to do more optimizations.

PR #304, opened 2026-04-30
wildbits: build sysgo from platform source

DrPitre:
## Summary
- make the shared WildBits recipe build  from  via the RMA/RLINK-style object path
- ensure Level 2 WildBits picks up  instead of the generic 
- keep the existing FEU-specific  override behavior aligned with the shared recipe approach

## Why
The shared WildBits recipe could satisfy  through the generic direct-assembly pattern rule, which let Make resolve  from . WildBits Level 2 should instead build from the platform-specific  source.

## Testing
- not run


MatthewMassie (approved):
Approved.

PR #305, opened 2026-05-01
Clean up bitmap and text block mixups

Nitrobotics:
View, scf, shellbg, text colors all appear to be in working order for Core1X
Please thoroughly test


MatthewMassie (approved):
Tested scfg, view, Basic09 with multiple programs, shellbg, shellbgoff, changed default startup with scfg.  Cursor looks fixed, etc. Approved.

PR #306, opened 2026-05-01
wildbits: fix shared Level 2 recipe command builds

DrPitre:
Summary
- add the missing Level 2 port include path to the shared WildBits recipe
- bring the WildBits Level 2-only command set into recipes/wildbits/wildbits.mak
- fix level2/wildbits/cmds/wild.asm so defsfile is loaded before os9.d

Details
The shared WildBits recipe was missing several commands that the dedicated WildBits Level 2 build used to include, and wild would fail to assemble because Level equ 2 was not defined before os9.d was read.

This change:
- adds -I$(L2PD) for Level 2 WildBits recipe builds
- adds the Level 2-only commands dmem, minted, mmap, modpatch, proc, pmap, smap, gfxstatus, xtclut, drawtest, play, wild, shellbg, shellbgoff, ntptime, and view
- leaves out cpm and rogue for now
- moves use defsfile ahead of use ../defs/os9.d in wild.asm so F$MapBlk and F$ClrBlk resolve correctly

Verification
- make -C recipes/wildbits/l2 PLATFORM=jr2

### DIFF
--- a/defs/wildbits_vtio.d
+++ b/defs/wildbits_vtio.d
@@ -170,7 +170,6 @@ V.LY                RMB      1                    line coordinate for Y
 V.GCADDR            RMB      3                    address of cursor on screen
 V.GCAD8K            RMB      2                    address in 8K window
 V.GMAPBLK           RMB      2                    mapped in logical address of block
-V.FONTPATH          FCC      \/dd/sys/fonts/\
 V.FONTNAME          RMB      33
 
 * Start of Line Interrupt Handling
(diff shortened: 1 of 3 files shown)


Nitrobotics (approved):

MatthewMassie (commented):
This is John's work. I see you removed the reference to font in wildbits_vtio.d (Understand why) but what resolves the path for programs using V.FONTPATH?

PR #309, opened 2026-05-02
wildbits: copy FEU startup into recipe disk images

DrPitre:
## Summary
- add FEU folder creation to the shared WildBits recipes
- generate a single FEU startup script for recipe-based disk images
- copy that script into FEU/startup for both Level 1 and Level 2 WildBits builds

## Background
The older WildBits makefiles populate a FEU directory in the generated disk images, but the shared recipes did not. That meant recipe-built WildBits images were missing the FEU startup entry point that the older build flow provided.

## What changed
This updates recipes/wildbits/wildbits.mak to:
- generate one shared FEU startup script during the recipe build
- add FEU as a disk image directory
- copy the generated script into FEU/startup
- clean up the generated startup artifact on clean

This intentionally uses one common startup file rather than the older media-specific FEU startup variants.

## Verification
- make -C recipes/wildbits/l1 -n PLATFORM=jr
- make -C recipes/wildbits/l2 -n PLATFORM=jr2

Both dry runs now show:
- os9 makdir ...,FEU
- os9 copy -o=0 -l feu.startup ...,FEU/startup

### DIFF
--- a/recipes/wildbits/wildbits.mak
+++ b/recipes/wildbits/wildbits.mak
@@ -74,6 +74,7 @@ endif
 BASIC09 = basic09 runb inkey syscall wild
 BASIC09_FILES = $(wildcard $(3RDPARTY)/packages/basic09/samples/*.b09)
 STARTUP = $(LEVEL2)/wildbits/startup
+FEU_STARTUP = feu.startup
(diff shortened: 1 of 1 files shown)


MatthewMassie (approved):
Tested. Now booting auto booting Level 2 from recipes. Approved.

PR #310, opened 2026-05-02
ident: skip non-module bytes while scanning

DrPitre:
## Summary
- make  continue scanning when a byte offset is not the start of an OS-9 module
- advance one byte at a time until a valid  header is found
- preserve normal module processing once a real header is encountered

## Why
When scanning blobs that contain padding, garbage, or other non-module bytes between modules,  should not stop at the first invalid header candidate. This change lets it skip over non-module bytes and keep searching for the next real module.

## Verification
- built  successfully with


MatthewMassie (approved):
Tested. Now able to see OS9BOOT modules.

PR #313, opened 2026-05-05
Enable noscroll status line

rlucente-retro:
**Summary of Changes**
This PR completes the implementation and validation of the vertical split-screen windowing system for the Wildbits VTIO driver. It addresses critical bugs in scrolling and coordinate mapping while providing a robust test suite for the new capabilities.

**Key Technical Fixes**
- SCROLL Routine: Corrected a stack addressing error (sta 1,s to sta ,s) that was corrupting the cursor column and causes out-of-bounds row erasures during scrolls.
- CurXY Logic: Hardened the cursor positioning routine with unsigned comparisons and underflow protection to ensure relative coordinates are strictly clamped within window boundaries.
- Boundary Safety: Ensured all physical row calculations (Relative + WStartY) result in valid screen offsets (0-29) to prevent memory corruption.

**Enhancements**
- Multi-Pane Support: Fully implemented DWSet (1B 20) for splitting the screen into upper (Window 0) and lower (Window 1) panes.
- State Preservation: Implemented DWSelect (1B 21) to correctly save and restore cursor position and color state when switching between active panes.
- FEU Startup Integration: Updated the First Execution Unit (FEU) build recipe to utilize a partitioned display, pinning the system header to a 5-line upper pane while keeping the lower pane active for user interaction.

**Validation & Testing**
- New Test Suite: Added test_upper and test_lower to the Level 2 Wildbits distribution. These scripts provide exhaustive coverage for edge cases, including 1-line windows, color persistence, and coordinate wrapping.
- Hardware Accuracy: Calibrated test scripts to match actual Wildbits hardware palette indices.
- Desk Checks: Verified all logic via step-by-step register/variable traces (documented in test_upper_final.csv and test_lower_final.csv).


jfed6000:
Does this need to be in the driver?  We should have a call to discuss this.  I'll check this out later today.  Is this just for one process or separate processes?  Is it creating a separate device for each window?

DrPitre:
This work may collide with the work that you're working on John. I'm happy to defer the PR for the moment.

jfed6000:
It does a bit, but doesn't 1B20 affect windows that are separate devices?  Is this just splitting a display for 1 device?  Is the point of this to just have a status bar? If so, would it be better to have a setting to turn off scrolling (if it doens't exist already) and let the program manage the screen on its own?  I don't want to discourage people from submitting PRs, I'm just not sure this is the best way to accomplish this.  I have a status line in te at the top and the bottom of the screen which works with the current driver.

jfed6000:
I just wonder if a better way wouldn't be to define the scrolling region as T lines from the top and B lines from the bottom, and then the user program can just use the cursor controls to change between the areas.  So T=0,B=0  would be whole screen scroll, T=1,B=1 would keep the scrolling area from line 2 to line 29 on a 30 line screen, and give you a top and bottom status bar.  This would work on any size screen.

rlucente-retro:
I really thought this PR was dead in favor of the other windowing solution and just having the status bar be a separate region. Right now, I'm using the escape sequence in Level 2 to delete a line at line one to force the text to scroll up. That way, the status line remains unchanged. My app just counts the number of lines sent to the screen and reacts appropriately.

jfed6000:
Cool!  Let's pause this then, unless you think we can just delete this.

rlucente-retro:
We can just delete this PR

jfed6000:
Closed PR Request

PR #316, opened 2026-05-08
wildbits: add DriveWire recipes

DrPitre:
Summary: add WildBits DriveWire recipe targets for Level 1 and Level 2, teach the shared WildBits recipe how to swap SD vs DriveWire boot composition and disk formatting, add local defsfile files in l1dw and l2dw so kernel builds resolve the intended Level-specific definitions, and remove the duplicate SCF terminal entry from the CoCo 3 DriveWire recipe.

Details: this introduces recipes/wildbits/l1dw and recipes/wildbits/l2dw and updates recipes/wildbits/wildbits.mak so recipes can override the format command, the primary boot RBF descriptor, and extra boot module groups. While validating l1dw, the kernel build failed because krn.asm starts with use defsfile and the new DW recipe directories did not yet provide the same local defsfile used by the existing l1 and l2 recipes. Adding those defsfile files fixes that and makes the new DW recipes build the expected Level 1 and Level 2 paths.

This branch also includes a small DriveWire cleanup in recipes/coco3/dw/recipe.mak: term_scdwv.dt was removed from the SCF list so the DriveWire terminal stack is not declared twice.

Verification:
- make -C recipes/wildbits/l1dw PLATFORM=jr
- make -C recipes/wildbits/l2dw -n PLATFORM=jr2 bootfile

### DIFF
--- a/recipes/coco3/coco3.mak
+++ b/recipes/coco3/coco3.mak
@@ -6,6 +6,20 @@ RECIPE ?= coco3
 -include recipe.mak
 vpath %.asm $(LEVEL1)/coco1/modules
 
+ifeq ($(CPU),6309)
+AFLAGS += -DH6309=1
+COCO3_LIB = libcoco3_6309.a
+NOS9_LIB = libnos96309l2.a
+COCO3_LFLAG = -lcoco3_6309
+else ifeq ($(CPU),6809)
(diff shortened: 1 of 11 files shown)


MatthewMassie (approved):
Tested DriveWire functionality with recipes.

PR #317, opened 2026-05-10
Add lldummy test driver

DrPitre:
Summary: add lldummy, a simple Level 1 low-level driver for testing, and bring the branch up to current main.

Details: this branch adds one new module at level1/modules/lldummy.asm. The goal is to provide a lightweight dummy low-level driver for test and development scenarios without changing any existing drivers.

Verification: the branch merged cleanly with current main, and the functional branch-only change remains the new lldummy driver module.


Nitrobotics (approved):
Builds fine.  Boot looks good.

PR #319, opened 2026-05-10
basic09: incorporate original Microware source comments, labels, and version history

DrPitre:
## Summary

- Merges block comments and inline (end-of-line) comments from the original Microware BASIC09 source files into the disassembly, adding context where none existed
- Replaces generated `Lxxxx` address labels with the original meaningful symbol names from the Microware sources (~1,000 labels replaced)
- Incorporates the original Microware edition history (editions 18–22, 1982–1983) into the file header alongside the existing NitrOS-9 change log
- Reformats the NitrOS-9 V1.21 change log entries to match the standard `Edt/Rev  YYYY/MM/DD  Author` header format used by the Microware entries

## Test plan

- [x] `basic09_6309` assembles byte-for-byte identically to main
- [x] `basic09_6809` assembles byte-for-byte identically to main


Wayne-Ca (approved):
I think these changes are necessary in order to understand exactly what the code is doing. I approve of this. My only question is, what of the labels that aren't found in the original source? Is there a way to correlate them back to the original?

Also, I would like to make a request, if possible. Since NitrOS-9 uses a shell that incorporates **CD** and **CX** as shortcuts for **CHD** and **CHX**, and since **PRINT** and **REM** have alternate shortcuts (**?** and **!** respectively), is it possible to add shortcuts to **CHD** and **CHX** so **CD** and **CX** can be used in Basic09?

Wayne-Ca:
Would a re-assembly of the original code produce a .asm output that would supply more comprehensive detail than this disassembly?

DrPitre:
Given the improvements that have been made since the original source code, it would be easier to bring in the comments in the disassembled version. There's more to do in this effort, but it's coming along.

DrPitre:
> I think these changes are necessary in order to understand exactly what the code is doing. I approve of this. My only question is, what of the labels that aren't found in the original source? Is there a way to correlate them back to the original?
> 
> Also, I would like to make a request, if possible. Since NitrOS-9 uses a shell that incorporates **CD** and **CX** as shortcuts for **CHD** and **CHX**, and since **PRINT** and **REM** have alternate shortcuts (**?** and **!** respectively), is it possible to add shortcuts to **CHD** and **CHX** so **CD** and **CX** can be used in Basic09?

That's doable. Feel free to make the change on a fork and create a pull request.

tlindner (approved):
md5 matches.

PR #321, opened 2026-05-10
basic09: remove dead runbt_6809 makefile target

DrPitre:
## Summary

- Removes the `runbt_6809` target which built `basic09.asm` with `-aRUNTIME=1`
- `basic09.asm` has no `RUNTIME` conditionals, so this target produced the same binary as `basic09_6809` — it was never useful
- Investigation confirmed that `runb.asm` and `basic09.asm` are independently disassembled binaries with largely different code; consolidation via a single source file is not feasible

## Test plan

- [x] `make basic09_6309 basic09_6809` still assembles cleanly


Wayne-Ca:
This change will just remove unnecessary code. I approve.

Wayne-Ca (approved):
Remove this code as it is wholly unnecessary.

PR #323, opened 2026-05-11
scripts: add one-space-rule normalizer and verification tooling

DrPitre:
## Summary

This PR adds two scripts that implement the **one-space rule** for NitrOS-9
assembly sources: a normalizer that collapses multi-space field separators to
single spaces, and a verification script that proves the transformation is
semantically neutral before it is committed.

---

## Background: single commit vs. history rewrite

Two approaches exist for applying a repo-wide style normalization.

### Option A — rewrite history with `git filter-repo`

Every historical commit gets the normalization applied.  The full record of
changes, authors, dates, and messages is **preserved** — only the SHA
fingerprints change.

| Pro | Con |
|-----|-----|
| `git blame` on any file shows meaningful changes, never dominated by a whitespace flood | All SHAs change — every contributor must `git fetch && git reset --hard origin/main` or re-clone |
| Consistent style throughout the entire history | External SHA references (issue trackers, release tags, CI pinned commits, third-party patches) break silently |
| Normalization is baked in from day one | Annotated tags must be force-deleted and recreated |

This is the right choice if the team can coordinate a hard cutover ("everyone
re-clones on date X") and is comfortable updating any external SHA references.

### Option B — single normalization commit (this PR's approach)

One commit on a feature branch applies the transformation to the working tree.
History before the commit is untouched.

| Pro | Con |
|-----|-----|
| All existing SHAs remain valid — `git pull` is all anyone needs | `git blame` on normalized files points to the normalization commit for most lines (use `git blame -w` to ignore whitespace) |
| No force-push of `main`; forks need no special handling | Pre-normalization history retains inconsistent spacing |
| Simpler to execute and verify | |

This PR implements Option B.  If the project later decides Option A is
preferable, `scripts/asm-normalize.py` can be fed directly to
`git filter-repo --force --blob-callback` to rewrite history instead.

---

## The one-space rule

Each logical field of an assembly line — label, opcode, operand, comment — is
separated by **exactly one space**.  Examples:

```
* Before                                  After
foo                 lda       #$FF   →   foo lda #$FF
                    fcs       /Init/ →    fcs /Init/
line020             fcc       %  use  /% →  line020 fcc %  use  /%
```

Rules the normalizer follows:

- Lines whose first non-whitespace character is `*`, `;`, or `#` are passed
  through unchanged (full-line comments).
- Blank lines are preserved.
- Operands of `fcc`, `fcs`, `fcn`, `fct`, and `fcs2` that start with a
  non-alphanumeric delimiter character have their **delimited content preserved
  verbatim** — internal spaces are part of the data.  The delimiter can be any
  character except an alphanumeric identifier start, `$`/`&`/`@` (numeric
  literal prefixes), or `%` followed by binary digits `[01]` (binary literal
  prefix).  A bare `%` used as a string delimiter (e.g. `fcc %hello world%`) is
  correctly recognized and preserved.

---

### `scripts/asm-normalize.py`

Normalizes one or more `.asm`/`.as` files **in-place** using an atomic
write (temp file + rename) so a failure never truncates the original.

```bash
# Normalize a single file
python3 scripts/asm-normalize.py level1/modules/init.asm

# Normalize all assembly sources in the repo
find . \( -name '*.as' -o -name '*.asm' \) ! -path '*/.git/*' \
    | xargs python3 scripts/asm-normalize.py

# Use as a stdin → stdout filter
python3 scripts/asm-normalize.py < foo.asm
```

### `scripts/verify-normalize.sh`

End-to-end verification that the normalizer produces **bit-identical OS-9
modules**.  Run this before committing the normalization to `feature/one-space-rule`.

```bash
bash scripts/verify-normalize.sh
```

What it does:

1. Clones `main` to `/tmp/nitros9_main_build` and builds it (detached HEAD so
   both builds embed the same `defs/buildinfo` string).
2. Creates the branch `feature/one-space-rule` off `main`, applies the
   normalizer to every `.asm`/`.as` file, and does a clean build.
3. Runs `os9 ident` on every built module in both trees and compares
   **name + size + CRC** for all modules except those built from sources that
   embed a wall-clock timestamp (`sysgo.asm` uses `dts`, `wbinfo.as` uses
   `dtb`); those are compared by **name + size only**.
4. Exits 0 on success, 1 on any mismatch.

**Pre-conditions:**
- `os9`, `lwasm`, and `python3` must be on `PATH`.
- The branch `feature/one-space-rule` must not already exist
  (`git branch -D feature/one-space-rule` to remove a stale run).
- Run from `main` or from this branch (`feature/normalize-scripts`).

## Test plan

- [x] `asm-normalize.py` passes 13 hand-written corner-case tests covering
  labels, opcodes, operands, inline comments, string delimiters (`/`, `"`,
  `!`, `%`), binary literals (`%01…`), comment-only lines, and blank lines
- [x] `verify-normalize.sh` completed a full dual build of the entire NitrOS-9
  tree with zero size or CRC differences across all 20,362 OS-9 modules
  (excluding the known `dts`/`dtb` timestamp modules, which matched on size)

### DIFF
--- a/scripts/asm-normalize.py
+++ b/scripts/asm-normalize.py
@@ -1,0 +1,133 @@
+#!/usr/bin/env python3
+"""asm-normalize: normalize 6809/6309 assembly source to one-space rule.
+
+Each field (label, opcode, operand, comment) is separated by exactly one
+space.  Content inside string directive delimiters is preserved verbatim.
+Lines whose first non-whitespace character is *, ;, or # are passed through
+unchanged, as are blank lines.
+
+Usage:
+  Single file (in-place):  asm-normalize.py file.asm [file2.asm ...]
+  Filter (stdin→stdout):   asm-normalize.py
+"""
+
+import os
+import re
+import sys
+
+# Directives whose operand is a delimited string; internal spaces must be kept.
+STRING_DIRECTIVES = frozenset({
+    'fcc', 'fcs', 'fcn', 'fct', 'fcs2',
+})
+
+# If the first operand character matches this pattern it is NOT a string
+# delimiter — fall back to plain token parsing instead.
+# % and $ are only non-delimiters when followed by their numeric digit sets
+# (%[01] = binary literal, $[0-9A-Fa-f] = hex literal); bare % or $ IS a
+# valid fcc/fcs delimiter and must be treated as such.
+_NOT_A_DELIM = re.compile(r'^(?:\w|[$&@]|%[01])')
+
+
+def _normalize_line(line: str) -> str:
+    # Preserve line ending
+    if line.endswith('\r\n'):
+        eol, line = '\r\n', line[:-2]
+    elif line.endswith('\n'):
+        eol, line = '\n', line[:-1]
+    else:
+        eol = ''
+
+    # Blank line
+    if not line.strip():
+        return eol
+
+    # Comment / preprocessor directive: *, ;, or # as first non-whitespace char
+    first_nws = line.lstrip()
+    if first_nws[0] in ('*', ';', '#'):
+        return line + eol
+
+    # --- Parse label -------------------------------------------------------
+    label = ''
+    if line[0] not in (' ', '\t'):
+        m = re.match(r'(\S+)[ \t]*(.*)', line)
+        if not m:
+            return line.rstrip() + eol     # label-only line
+        label, rest = m.group(1), m.group(2)
+    else:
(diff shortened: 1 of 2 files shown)


DrPitre:
Closing because the repository has standardized on scripts/asmprettyprint.py and its pre-commit wrapper. PR #375 removed the redundant formatter and filter stack, so adding a second normalization implementation would reintroduce competing formatting paths.

PR #326, opened 2026-05-13
Wildbits: View enhancements

Nitrobotics:
Default is bitmap #0
-#  set and view bitmap 0-2
-r  free all bitmaps and return to text
{filename} load and view

### DIFF
--- a/level1/wildbits/cmds/view.asm
+++ b/level1/wildbits/cmds/view.asm
@@ -13,6 +13,10 @@
 *
 *   3      2025/11/08  Roger Taylor
 * Add option to show a specific bitmap # if no file is specified.
+*
+*   4      2025/05/09  Roger Taylor
+* Improve pixel block calculation speed in GetXYBlk routine
+* Added -r option to free all bitmaps and return to text mode, exit
 
                     nam       view
                     ttl       Picture viewer
@@ -26,7 +30,7 @@
 tylg                set       Prgrm+Objct
 atrv                set       ReEnt+rev
 rev                 set       $00
-edition             set       1
+edition             set       4
 
                     mod       eom,name,tylg,atrv,start,size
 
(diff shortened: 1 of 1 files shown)


level1/wildbits/cmds/view.asm
```diff
@@ -98,40 +103,60 @@ inittext            fcc       /Picture Viewer by Roger Taylor/
 
 clutpathname        fcn       "/dd/cmds/xtclut"
 
+skipspace           lda       ,x
```
DrPitre: One minor suggestion: flip the comparison and save an instruction:

```
skipspace lda ,x+
 cmpa #C$SPAC
 beq skipspace
 leax -1,x
 rts
```

level1/wildbits/cmds/view.asm
```diff
@@ -98,40 +103,60 @@ inittext            fcc       /Picture Viewer by Roger Taylor/
...
+                    std       <fmemsize
+                    clr       <filepath
 
                     clra
                     clrb
-                    std       bitmapnum
-                    std       clutnum
+                    std       <bitmapnum
+                    std       <clutnum
                     lda       #1                  Path #
```
DrPitre: Minor optimization: `lda #1` ➞ `inca` (saves one byte)

DrPitre (commented):
Thanks for this PR! I've made a few suggestions.

DrPitre (changes requested):

level1/wildbits/cmds/view.asm
Nitrobotics: Ok, I put back in the 'space' routine from merge.asm which I originally had.  Not sure why I bloated that routine later.

level1/wildbits/cmds/view.asm
Nitrobotics: clra works as long as Path is always 1.  I've pushed this change.  Thanks.

DrPitre (approved):

PR #327, opened 2026-05-13
Remove docs folder in favor of wiki

DrPitre:
## Summary
- remove the entire tracked docs/ tree from the repository
- treat the project wiki as the canonical home for documentation going forward
- stop carrying duplicate documentation sources and generated assets in the repo

## Rationale
The documentation is now maintained in the wiki, so the in-repo docs folder is redundant and can be removed to reduce maintenance overhead.


Wayne-Ca (approved):
This makes sense as the use of a wiki allows everyone to view and read the documentation, not just those in this repo.

PR #333, opened 2026-05-17
Reintroduce the cmoc_os9 CoCo 3 recipe

DrPitre:
## Summary
- reintroduce `recipes/coco3/cmoc_os9/` on top of `main`
- delegate recipe settings to the authoritative `cmoc_os9` sibling repository
- build CoCo 3 graphics tests from `graphictest/`, headless tests from `unittest/`, and utility programs from `utils/`
- make CoCo 3 keyboard repeat defaults configurable for the `cmoc_os9` recipe path

## Details
- restores the `defsfile`, `makefile`, and `recipe.mak` entrypoints removed from `main`
- keeps `startup` and `test` sourced from `cmoc_os9/recipes/coco3/cmoc_os9/` instead of reviving stale copies in `nitros9`
- preserves the recipe-specific disk-image rule that copies both the startup script and test script onto the disk image
- adds `CMOC_OS9_UTILS_DIR` and corresponding build/copy rules so the recipe can include command programs from `cmoc_os9/utils`
- replaces the hard-coded CoCo 3 INIT key repeat constants with `INIT_KEYRPTSTART` and `INIT_KEYRPTDELAY`, which still default to the current values unless overridden at build time
- this key repeat configurability was added specifically to make CoCo 3 keyboard behavior easier to tune when running under MESS with `-nothrottle`

## Notes
- this PR intentionally reintroduces the deleted recipe path because `main` no longer contains it
- the recipe expects a sibling `cmoc_os9` checkout, consistent with the current split of source ownership
- the generated `level2/coco3/modules/init` build artifact is not included in the change


Wayne-Ca (commented):
This seems to be a valid change. I'm not really qualified to speak on it, but it makes sense to me.

PR #334, opened 2026-05-21
Annotate and shrink the more command

DrPitre:
## Overview

This updates `level1/cmds/more.asm` with a cleaner OS-9 assembly presentation while preserving the optimized command size.

The source now has an OS-9-style `Edt/Rev` history block, lowercase inline comments, and more complete annotation for storage, data tables, option parsing, paging, prompt handling, EOF handling, and exit paths.

## Size

| Build | Size |
| --- | ---: |
| Original `more` | 434 bytes |
| Updated `more` | 413 bytes |
| Savings | 21 bytes |

## Notable Changes

- Adds an `Edt/Rev` history section to the source header.
- Documents direct-page variables and embedded prompt/header data.
- Annotates each executable instruction with lowercase line comments.
- Keeps the compact layout using the existing size optimizations.
- Leaves unrelated worktree changes out of the commit.

## Verification

```sh
python3 scripts/asmprettyprint.py level1/cmds/more.asm
(cd level1/coco2/cmds && make -B more && wc -c more)
```

Verified output size:

```text
413 more
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,9 @@ buildinfo
 bootfile
 kerneltrack
 *.dsk
+
+# Third-party package scratch files
+3rdparty/packages/ed/c.errors
+3rdparty/packages/uemacs/test
+3rdparty/packages/uemacs/test.bin
+3rdparty/packages/uemacs/test.c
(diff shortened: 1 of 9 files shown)


Wayne-Ca (commented):
This looks good. Anything that reduces file size is a win, especially if it retains the module size.

PR #339, opened 2026-05-22
SAVEM and LOADM commands added

Wayne-Ca:
SAVEM allows the user to save unpacked procedure modules to an external file. This allows preservation of the original source, including all comments, unused line numbers and all variable names. They can be saved individually or as merged module files. The modules are not packed, and have no CRC as well as the header parity and procedure status bytes are not set. Ident will show a bad CRC. The modules are saved to files using a .up extension (means unpacked).

LOADM allows the user to load unpacked procedure files (files with a .up extension) into Basic09's workspace.

These commands enable the user to preserve their code in the event the source code files are lost.


DrPitre:
Wayne, this brings in a bunch of non-related changes. We need to meet so you can understand how to properly stage commits. I'm closing this PR.

Wayne-Ca:
I tested both 6809 and 6309 versions in VCC using NitrOS-9 EOU. Hardware and other emulator testing would be appreciated.

PR #341, opened 2026-05-22
Feature/savem loadm

Wayne-Ca:
SAVEM and LOADM allow the user to save and reload unpacked modules in the Basic09 workspace. Files created with SAVEM have a .up extension. Modules can be saved individually or as merged module files.

LOADM looks for files with the .up extension and loads them if found.

Tested 6809 and 6309 versions in VCC using NitrOS-9 EOU. Hardware testing and other emulator testing is needed.


Wayne-Ca:
This is a modification to Basic09.

PR #342, opened 2026-05-22
basic09: add SAVEM and LOADM commands

Wayne-Ca:
## Summary
- Add `SAVEM` command to save unpacked Basic09 modules to disk with a `.up` extension
- Add `LOADM` command to load unpacked modules back into the Basic09 workspace

## Test plan
- [ ] Build basic09 and verify it assembles without errors
- [ ] Test SAVEM saves a module file with `.up` extension
- [ ] Test LOADM reloads a previously saved module
- [ ] Verify existing Basic09 commands are unaffected

- There are times when even the best programmers can lose their source code. Traditionally, where Basic09 is concerned, the only choice was to start again and rewrite the code from scratch. This usually resulted in the module being different from the original unless you tweaked it until it matched.
- I wrote DCom, then decode and finally Unpack to alleviate this problem. Then I thought, why can't it be possible to save what's in the Basic09 workspace? That would preserve all of the elements of the source code and make restoring the code if you lose the source easier and result in the same module when packed.
- This is what SAVEM and LOADM are designed to do. With SAVEM you can save your unpacked modules to files using a .up extension to make them easier to identify. With LOADM you can load those .up files back into Basic09's workspace. Once reloaded you can SAVE them back out as source code files. The resulting source code file(s) will be identical to the source code file you worked so hard to write.
- Best of all, you don't need to learn new parameters for saving and loading. The syntax is the same as for SAVE and LOAD. You're just adding a M to the command, so SAVEM instead of SAVE and LOADM instead of LOAD.


DrPitre (commented):
Wayne, claude settings.local.json and the basic09 binary and SAVEM_syntax.md are still leaking into your commit. I would suggest carefully studying what you are committing before doing so.

DrPitre:
Wayne, there is still the lingering .claude folder and settings. You need to be more rigorous and thorough when committing these.

Wayne-Ca:
I finally have it where the .md file and the binary are not included in the PR. It took some doing, but finally!

DrPitre:
Wayne,

The .claude folder and the settings file is still lingering.

As I stated, you need to be rigorous about cleaning up these unnecessary files. It's causing unnecessary churn for us.

Wayne-Ca:
The json file is gone. I'm learning. Claude is a great help, but it doesn't seem to know what should and shouldn't be included in a PR.

Wayne-Ca:
Everyone: Sorry for the mess. It was unintentional. I'll do better in the future.

DrPitre:
Ok, now that this is in better shape...

Please explain in your PR what this does and why it's needed.

DrPitre:
Closing here because BASIC09 source has moved out of this repository. This feature remains worth considering, but it should be rebased onto the Microware source of truth and proposed in nitros9project/nitros9-languages.

strickyak:
Do we allow feature creep into the Basic09 main tree?  

Basic09 fits in specific historic Level 1 environments and is pretty big already, 
but if it grows, some use cases will not work.

I think we should maintain Basic09 in its current size, but you can start a bigger one under a new name
as an option, and add all the features you want.

Wayne-Ca:
> Do we allow feature creep into the Basic09 main tree?
> 
> Basic09 fits in specific historic Level 1 environments and is pretty big already, but if it grows, some use cases will not work.
> 
> I think we should maintain Basic09 in its current size, but you can start a bigger one under a new name as an option, and add all the features you want.

This is a feature I wanted to have back in 1990. DCom, decode and Unpack would be totally unnecessary if they had existed. However, I get your point. I only made this change to the Level 2 Basic09. The 6809 version is slightly over 24K and the 6309 version is slightly under 24K. I would not add this to the Level 1 versions because the Level 1 version of the OS only supports 64K max and the system uses it as well. There wouldn't be enough memory left to do anything. That said, Dr. Pitre has moved Basic09 to a different repo, and I will see about working from there. I will take your advice see about making my edits a different version of Basic09 for Level 2. In fact, I was thinking I could make it a separate binary altogether so people can use it for the SAVEM/LOADM feature and not worry so much about memory space for editing/debugging and execution.

strickyak:
Thanks for the explanation, Wayne.
I would still encourage a new version that's open to new features.
Great things could happen, without breaking older patterns. 
I can recall doing things on a 128K coco3 under level2 with no 8k RAM chunk free. 
Retro is all about tradeoffs (:

PR #348, opened 2026-05-28
Fix DoPoll carry leak masking IRQs after non-clock interrupt

DrPitre:
## Summary

- When a non-clock IRQ fires, `SvcIRQ` sets `D.SvcIRQ = DoPoll` and chains to `S.SysIRQ`. `DoPoll` exits its polling loop with carry set (its normal "no more IRQs" exit) and falls through to `DoToggle`, returning to `DoneIRQ` with carry set.
- `DoneIRQ` (`krn.asm:1036`) treats carry set as an error and ORs `IntMasks` into the saved CC on the interrupt stack. The subsequent `RTI` returns the interrupted process with I=1 (interrupts masked).
- With the 6809 I bit set, the 60 Hz GIME clock IRQ cannot fire. `D.Tick` never decrements, and `D.Year`/`D.Month`/`D.Day`/`D.Hour`/`D.Min`/`D.Sec` stop updating. A spinlocking process (no OS9 calls) will freeze the clock globals indefinitely.
- Fix: add `andcc #^Carry` in `DoToggle` (`clock.asm`) before its `rts`. Since `DoPoll` falls through to `DoToggle`, this covers the non-clock IRQ path. The clock IRQ path is unaffected — `SvcVIRQ` calls `DoToggle` as a subroutine but exits via `jmp [D.Clock]`, so carry state on return is irrelevant.
- Affects both **CoCo3** and **Wildbits** ports — the `NoClock`/`DoToggle`/`DoneIRQ` path is shared common code in both builds.

## Test plan

- [ ] Build Level 2 for CoCo3 and confirm no assembly errors
- [ ] Build Level 2 for Wildbits and confirm no assembly errors
- [ ] Boot NitrOS-9 Level 2 on CoCo3 (real or emulated), perform disk I/O, and verify that `D.Sec`/`D.Min` continue advancing after the disk IRQ returns
- [ ] Run a tight busy-loop process alongside disk activity and confirm clock globals (`F$Time`) still update correctly


DrPitre:
Closing — branch was cut from the wrong base. Will reopen from main.

PR #354, opened 2026-06-07
Ignore in-tree build output via per-directory .gitignore files

jamieleecho:
## First nitros9 commit

Hi, this is my first nitros9 commit. It was generated by Claude and tested by hand. The goal of the PR is to update the gitignores to make contributions a bit easier - knowing quickly what was changed and what should be checked in.

I did consider the alternative approach - build everything into a separate BUILD location - this would be preferable long term but likely very disruptive initially.

## Problem

A fresh `make` leaves ~5,000 untracked build artifacts. NitrOS-9 compiles modules directly into the source tree (assembled modules, boot files, ROM/disk images, compiled command binaries land beside their source), and **no directory is pure output** — so neither a single extension list nor a broad directory ignore is both safe and minimal.

## Approach

Each build directory's makefile already enumerates exactly what it produces in its `clean` target. So derive the ignore lists from `make` itself:

```sh
make -n clean MAKE=:
```

- `-n` prints the `rm -f …` list **without building or deleting** anything.
- `MAKE=:` neutralizes the recursive `$(MAKE) -C subdir clean` lines (GNU make runs those even under `-n`), so each directory reports **only its own** outputs — children write their own `.gitignore`.

`scripts/gen-gitignores.sh` writes one **anchored** `.gitignore` per directory from that output and is re-runnable when makefiles change.

## Why anchored entries

`make clean` only removes files in its own directory, so every entry is anchored (`/foo`). This prevents a glob like `/*.a` from shadowing tracked source in a subdirectory (e.g. `lib/kreiderclib/qsort.a`).

## Root negations

Seven checked-in deliverables (prebuilt `.dsk` images under `l2upgrade`/`os9l2bbs` and a sample `bin.list`, none of which have a build rule) were already being hidden by the existing broad `*.dsk`/`*.list` globs in the root `.gitignore`. Added explicit negations so they stay tracked.

## Result

- `git status` is clean after a build.
- Only real build products are ignored.
- `git ls-files --cached --ignored --exclude-standard` is empty — **no tracked file is shadowed**.

271 generated `.gitignore` files + the generator script + root `.gitignore` negations.

### With a fully built distribution, git status is empty
```
make clean && make && make dsk && git status
...
On branch gitignore-build-artifacts
Your branch is up to date with 'origin/gitignore-build-artifacts'.

nothing to commit, working tree clean
```

### With a fully built distribution, git shows no shadowed files
```
git ls-files --cached --ignored --exclude-standard 
jcho@Jamies-MacBook-Pro nitros9 % 
```


DrPitre:
I see several issues with this approach:

- Having a `.gitignore` in every folder seems way over the top. One `.gitignore` at the root of the project should be enough.
- We are moving away from building within the source folders to the `nitros9/recipes` folder, so mixing of object files and source files is no longer an issue.

PR #362, opened 2026-06-28
Implement Delete Line (`$1F $31`) Escape Sequence in Wildbits `vtio` Driver

rlucente-retro:
## Summary
This pull request implements the `$1F $31` (Delete Line) escape sequence in the NitrOS-9 Level 2 video driver (`vtio.asm`) for the the Wildbits platform. This change aligns the Wildbits video driver's behavior with the Tandy Color Computer 3's window driver (`cowin.asm` / `grfdrv.asm`), enabling text-mode applications that rely on partial-screen scrolling via Delete Line (such as the Infocom game port) to scroll correctly on the Wildbits platform.

---

## The Problem
Applications running on NitrOS-9 Level 2 often use partial-screen scrolling to scroll only the content area of the screen while keeping status lines (typically row 0) intact. On the CoCo 3, this is achieved by moving the cursor to row 1 (the line below the status line) and sending the escape sequence `$1F $31` (Delete Line). The CoCo 3 driver deletes row 1, which shifts all rows below it up by one row.

Prior to this change, the Wildbits `vtio.asm` driver only supported the `$20` (reverse video on) and `$21` (reverse video off) parameters for the `$1F` escape code prefix. The `$31` parameter was unhandled and discarded: the driver would immediately reset the escape handler to the default state and return without modifying the screen, causing the display to hang or fail to scroll.

---

## Solution & Technical Details

1. **Escape Character Routing**
   - Modified `OneEffHandler2` in [level1/wildbits/modules/vtio.asm](file:///Users/richardlucente/development/git/nitros9/level1/wildbits/modules/vtio.asm) to recognize the `$31` parameter.
   - Wrapped the routing instruction in a conditional assembly block (`ifgt Level-1`) so that the Delete Line sequence is only built for NitrOS-9 Level 2 configurations. If Level 1 is compiled, the sequence remains ignored as before.

2. **Delete Line Logic (`DeleteLine` Routine)**
   - **Corner Case: Bottom Row**: Compares the current cursor row (`V.CurRow`) against the screen height (`V.WHeight`). If the cursor is already at the bottom-most row, it skips scrolling and simply erases the bottom row with spaces (behaving identically to the CoCo 3 driver).
   - **Scroll Copy**: If the cursor is above the bottom row, it calculates the number of rows below the cursor to scroll up.
   - **MMU Page Mapping**: Disables interrupts and maps page `$C2` (text RAM) and `$C3` (attribute RAM) in turn. It performs a block copy of both characters and attributes from the line below (`X + WWidth`) to the current line (`X`) in 2-byte increments.
   - **Last Line Cleanup**: Erases the bottom-most row of the screen (`V.WHeight - 1`) with background-colored spaces using the internal `EraseLineCore` helper.
   - **Restoration**: Restores the original MMU slot mapping, re-enables interrupts, and branches back to `ResetHandler` to reset the escape handler state.

3. **Robustness & Register Preservation**
   - **Y Register Preservation**: Since the scrolling loop uses register `Y` as a byte-copy counter, `Y` is pushed to the stack on entry to `DeleteLine` and popped prior to exit. This preserves the path descriptor address stored in `Y` and prevents register corruption in caller contexts.
   - **Assembler Compatibility**: Used a standard global label `dl_loop` for the scrolling loop to avoid local symbol resolution issues (`Undefined symbol`) under `lwasm` during Level 2 builds.

---

### 1. Build Verification
Compiled the release for the Wildbits JR2 Level 2 platform successfully:
```bash
export NITROS9DIR=$(pwd)
make -C recipes/wildbits/l2 PLATFORM=jr2
```
The driver compiled and assembled cleanly without any syntax errors, warnings, or byte overflows. The final release disk (`l2_wildbitsjr2.dsk`) built successfully.

### 2. Desk-Checking Traces
Performed step-by-step register/stack checking for three key positions (on a 24-row, 80-column screen configuration):
- **Bottom row (`CurRow = 23`)**: Skips copy-scrolling, restores stack/Y, erases row 23, and exits.
- **Top row (`CurRow = 0`)**: Copies rows 1–23 to rows 0–22, restores stack/Y, and erases row 23.
- **Next-to-last row (`CurRow = 22`)**: Copies row 23 to row 22, restores stack/Y, and erases row 23.
All paths maintain perfect stack alignment and correctly restore MMU mapping, interrupt masking state, and the `Y` register.


level1/wildbits/modules/vtio.asm
```diff
@@ -969,6 +969,10 @@ OneEffHandler2
                     beq       revon
                     cmpa      #$21
                     beq       revoff
+                    cmpa      #$31
+                    ifgt      Level-1
```
DrPitre: Is there a reason this should NOT work under Level 1?

level1/wildbits/modules/vtio.asm
```diff
@@ -2161,6 +2165,68 @@ Blk2Addr            clra                          clear a, block # is in b
                     lslb                          x32 ($20)
                     rola
                     rts
+
+DeleteLine
+* Preserve register Y (holds the path descriptor pointer for the active I/O path)
+                    pshs      y
```
DrPitre: I think we can defer this `pshs y` until line 2180 and then remove the `puls y` on line 2228 to save two bytes.

level1/wildbits/modules/vtio.asm
```diff
@@ -2161,6 +2165,68 @@ Blk2Addr            clra                          clear a, block # is in b
...
+* Check if cursor is at the bottom row (CurRow + 1 >= WHeight)
+                    lda       V.CurRow,u
+                    adda      #1
```
DrPitre: replace with `inca` to save one byte.

DrPitre (changes requested):
There's an opportunity to hand-code some of this assembler to make it tighter.

level1/wildbits/modules/vtio.asm
rlucente-retro: * **Analysis**: The Wildbits hardware platform handles text screen memory pages ($C2 for text characters, $C3 for character attributes) using a hardware Memory Management Unit (MMU) mapped page mechanism (`MAPSLOT`). This MMU mapping is defined and used for both Level 1 and Level 2 builds of `vtio.asm` (with Level 1 mapping to `MMU_SLOT_7` and Level 2 mapping to `MMU_SLOT_2`).
* **Implementation**:
  - Remove the conditional assembly blocks `ifgt Level-1` / `endc` restricting routing in [OneEffHandler2](file:///Users/richardlucente/development/git/nitros9/level1/wildbits/modules/vtio.asm#L967-L977).
  - Move the definition of [DeleteLine](file:///Users/richardlucente/development/git/nitros9/level1/wildbits/modules/vtio.asm#L2169-L2230) outside of the large `ifgt Level-1` conditional assembly block (placing it after the `endc` at line 2230).
  - Resolved in [18fd60b](https://github.com/nitros9project/nitros9/pull/362/changes/18fd60bb5d637eeac93dcca99d6bc5f6fc9b0f8d)

level1/wildbits/modules/vtio.asm
rlucente-retro: **Analysis**: On entry to `DeleteLine`, the `Y` register points to the active path descriptor. If the cursor is already at the bottom-most row, copying is skipped and we immediately clear the bottom row. Register `Y` is not modified by the bottom row check (`CurRow + 1 >= WHeight`). By deferring `pshs y` until we enter the copy-scrolling path:
  - We can branch directly to `dl_clear` using `bhs dl_clear` without needing to pop `Y`.
  - The entire cleanup label `dl_clear_y` and its associated `puls y` / `bra dl_clear` instructions can be deleted.
  - This saves **4 bytes** of code size and several execution cycles in the quick-exit path.
  - Resolved in [18fd60b](https://github.com/nitros9project/nitros9/pull/362/changes/18fd60bb5d637eeac93dcca99d6bc5f6fc9b0f8d)

level1/wildbits/modules/vtio.asm
rlucente-retro: Not only the `inca` change but also ...
**Analysis**: The original code computes `rows = WHeight - (CurRow + 1)` using stack pushing and pulling:
  ```assembly
  pshs      a                   * Push CurRow + 1 to stack
  lda       V.WHeight,u
  suba      ,s+                 * A = WHeight - (CurRow + 1), pops stack
  ```
  This requires **6 bytes** and **15 CPU cycles**.
  Since register `A` contains `CurRow + 1` right after the bottom-row check, we can negate it and add `V.WHeight,u`:
  ```assembly
  nega                          * A = -(CurRow + 1)
  adda      V.WHeight,u         * A = WHeight - (CurRow + 1)
  ```
  This register-only alternative requires only **3 bytes** and **6 CPU cycles**, saving **3 bytes** and **9 CPU cycles**.
Resolved in [18fd60b](https://github.com/nitros9project/nitros9/pull/362/changes/18fd60bb5d637eeac93dcca99d6bc5f6fc9b0f8d)

DrPitre (approved):

PR #386, opened 2026-08-14
  wildbits/vtio: fix cursor synchronization and remove write-ahead attribute bleed

rlucente-retro:
#### Summary of Changes

  1. Removed Speculative Write-Ahead in RawWrite:
      • Removed sta 1,x (introduced in 88742ef0).
      • Characters and attributes are written strictly to the active cell (,x), eliminating line-wrap attribute corruption across the right margin (Row 0 Column 79 → Row 1 Column 0) and avoiding stuck reverse video on mode changes (REVOFF).
  2. Hardware Cursor Color Synchronization in Write:
      • Updates Vicky II's dedicated hardware cursor color register (VKY_TXT_CURSOR_COLR_REG) with V.FBCol,u alongside VKY_TXT_CURSOR_X/Y_REG_L.
      • Keeps the hardware cursor in sync with the active text color in hardware without modifying underlying VRAM attributes or corrupting existing text when moving the cursor.
      • Clears the 6809 Carry flag on exit (andcc #^Carry) to prevent spurious OS-9 I/O driver errors.
  3. Safe Destructive Backspace in EraseChar:
      • Wrapped EraseChar with explicit MMU mapping for text ($C2) and attributes ($C3) under interrupt masking to ensure backspaced cells are cleanly cleared and colored.
  4. Branch Distance Fixes:
      • Updated branch targets to ResetHandler from short (bra/bne) to long (lbra/lbne) to eliminate assembly byte-overflow errors.


rlucente-retro:
changes are updated to the following

```
    ### wildbits/vtio: Row-Guarded Cursor Attribute Synchronization

    - **Row-Guarded Write-Ahead (`RawWrite`)**: Confines attribute lookahead strictly
  within the active row (`CurCol + 1 < WWidth`). When typing at Column 79, write-ahead
  is skipped so attributes never bleed into Row 1 Column 0 on line wraps.
    - **Clean Cursor Positioning (`Write`)**: Cursor positioning (`$02`, Home, Arrows)
  updates Vicky II hardware cursor registers and clears the Carry flag (`andcc
  #^Carry`) without modifying VRAM attributes, preserving status lines and existing
  text.
    - **Safe Destructive Backspace (`EraseChar`)**: Properly maps `$C2` to clear
  characters and `$C3` to restore the active background color under interrupt masking,
  keeping the cursor color consistent when backspacing.
```

rlucente-retro:
Here's a simple script to recreate the issue. Just drop this into SCRIPTS and then run SCRIPTS/testme.txt

The reason this happens is that the infocom interpreter calls ZUSL to update the status line. ZUSL only writes characters to the top line but because of attribute bleed over, an update to the status line will also cause the first character of normal game text on the second line to have reverse video attributes. This only happens on the wildbits platform and not Coco 3.

ZUSL can be called multiple times during the game. Whenever a new location is is entered (of course) but also when you issue multiple commands at once like "n.climb tree.get egg.d.s.e" the screen scroll will also trigger a status line update after the "[more]" prompt is displayed.

[testme.txt](https://github.com/user-attachments/files/31534027/testme.txt)

PR #390, opened 2026-08-20
K2/freeze fixes

Nitrobotics:
See the .md reports

### DIFF
--- a/level2/modules/clock.asm
+++ b/level2/modules/clock.asm
@@ -817,8 +817,12 @@ InitCont
                sta       INT_PENDING_0
                sta       INT_PENDING_1
                sta       INT_MASK_1
-               lda       INT_MASK_0
-               anda      #^INT_VKY_SOF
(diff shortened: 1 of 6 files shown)


recipes/wildbits/wildbits.mak
```diff
@@ -68,7 +68,10 @@ BOOTMODS = krn krnp2 ioman init \
 	$(BOOTMODS_EXTRA)
 endif
 
-SHELLMODS = shellplus date deiniz echo iniz link load save unlink
```
DrPitre, line 71: I think this is a red herring. Can we not include this change after some testing?

level2/modules/clock.asm
```diff
@@ -817,8 +817,12 @@ InitCont
                sta       INT_PENDING_0
                sta       INT_PENDING_1
                sta       INT_MASK_1
-               lda       INT_MASK_0
```
DrPitre, line 820: I'm not confident that this code is necessary.

PR #391, opened 2026-08-21
Fix stack bug and debug mode in wildbits rbmem

Nitrobotics:
Read from /f0, read/write to /f1 on the Jr2 and K2.
This includes format /f1, writing files, reformatting, etc.
Since /f0 is used for FEU we have it marked as a Read-Only drive.
You could change /f0 to r/w in your OS-9 build and make changes to FEU/f0 at run-time.


jfed6000 (approved):
Approved

PR #394, opened 2026-08-26
Add K2 support to WizFi

Nitrobotics:
Requires the new K2 core (wildbits_k2_6809_int_wizfi_20260825)
Fixes a few WizFi convenience scripts.
Fixes the modem command's -t0 mode.
Reworked WizFi driver.


MatthewMassie (approved):
Updated FPGA firmware and build PR 394 and tested WizFi device.

- Connects OK
- echo AT>/wz  OK
- ntptime -4 OK

Approved. Nice work Roger!

PR #395, opened 2026-08-27
Fix wildbits l2 drivewire boot

rlucente-retro:
### Summary
Streamlines recipes/wildbits/l2dw/recipe.mak to prevent DAT mapping corruption during early Level 2 kernel initialization.

### Problem
The l2dw recipe previously packed non-essential modules (scdwv, n–n5, pipeman, sc16550) into OS9Boot, inflating the bootfile to 32,512 bytes (5 physical 8KB blocks). During early kernel boot in krn.asm, F$SRqMem allocated 5 blocks for the bootfile, displacing CPU Slot 7 ($E000–$FFFF) from physical Block 7 to Block 6. This caused the CPU to execute font bitmap data instead of the kernel at $EE00, halting boot before VTIO initialization.

### Solution
• Removed non-essential modules from BOOTMODS_EXTRA and SCF_EXTRA in recipes/wildbits/l2dw/recipe.mak so only essential bootstrap drivers (dwio_serial, rbdw, ddx0) are embedded in OS9Boot.
• Reduced bootfile size from 32.5 KB to 28.6 KB (under the 32 KB / 4-block threshold). Secondary modules remain available on the /DD disk image for on-demand dynamic loading.

### Testing
• Verified clean build of l2_wildbits_dwjr2.dsk.
• Tested DriveWire bootstrap via pyDriveWire in MAME; NitrOS-9 Level 2 boots cleanly from /x0/OS9Boot, opens /Term, and reaches the Shell+ prompt on /DD.
• Tested DriveWire bootstrap via pyDriveWire and USB-C connection to Wildbits Jr2; NitrOS-9 Level 2 boots cleanly from /x0/OS9Boot, opens /Term, and reaches the Shell+ prompt on /DD.


rlucente-retro:
Specific steps to recreate this issue and why it's happening.
Create `~/.pydrivewirerc` with the contents:
```
option accept True
option port 65504
option uiPort 6800
option debug 1
dw disk insert 0 /Users/richardlucente/development/git/nitros9-main/recipes/wildbits/l2dw/l2_wildbits_dwjr2.dsk
```
Run pyDriveWire using `/path/to/pyDriveWire -d`
Run the [Wildbits Jr2 emulator](https://github.com/rlucente-retro/mame) using `/path/to/mame/mame wbjr2 -window`
When the FEU pick menu appears, select `o (Boot OS-9)` and then `x (Boot OS-9 from DriveWire)`

You'll see FEU locate /x0/OS9Boot and successfully transfer all 127 sectors (32,512 bytes) via DriveWire. Immediately after completing the transfer, the machine freezes with the following text:

        Searching for /x0/OS9Boot.
        /x0/OS9Boot found.
        Loading sector..................................................................
        .............................................................

Tracing hardware MMU registers and CPU program counter with the emulator confirms the bootfile spills over into physical block 3 which causes a one-off block shift of memory, replacing kernel code with font bitmap data and a subsequent crash.

Because the top 512 bytes of the address space ($FE00–$FFFF) are reserved for hardware I/O and vector space, the bootfile is loaded to end at $FDFF (immediately below $FE00).

Consequently, within the 32 KB window provided by physical Blocks 4 through 7 ($8000–$FFFF), only the range $8000–$FDFF is available to hold the bootfile:

    $FE00 - $8000 = $7E00  bytes  = 𝟑𝟐,𝟐𝟓𝟔  bytes (31.5 KB)

Any bootfile larger than 32,256 bytes ($7E00) forces the start address to drop below $8000 into physical Block 3 ($6000–$7FFF).

When the oversized bootfile spills into physical Block 3, early kernel initialization begins allocating physical memory starting at Block 3 instead of Block 4. This creates a one-block shift across all memory slots, erroneously remapping CPU Slot 7 ($E000–$FFFF) to physical Block 6 instead of physical Block 7. Because the CPU is actively executing kernel code in Slot 7, its instructions are instantly replaced by font bitmap data, causing a crash before initialization can complete.

This PR suggests changes to the bootfile, but there are obviously other combinations of descriptors/drivers/etc to stay below the 32,256 byte threshold.

rlucente-retro:
Also confirmed drivewire boot freeze on physical wildbits jr2 with macOS pyDriveWire with `~/.pydrivewirerc`:
```
option port /dev/cu.usbserial-600SA20462
option speed 230400

option uiPort 6800
option debug 1

dw disk insert 0 /Users/richardlucente/development/git/nitros9-main/recipes/wildbits/l2dw/l2_wildbits_dwjr2.dsk
```
Same behavior as emulator.

rlucente-retro:
Latest commit restores pipeman, piper, pipe and scdwv, n, n1–n5 to the Level 2 DriveWire bootfile, keeping OS9Boot at 30,720 bytes padded (under the 32,256-byte limit with 1,536 bytes of headroom). Unbundled drivers and descriptors are in /DD/CMDS on the disk image.

### In-Memory Modules

• Kernel & System: KrnP2, IOMan, Init, Clock, Clock2, Krn
• Console & Input: SCF, vtio, keydrv, TERM, font, palette, mousedrv
• Pipes: PipeMan, Piper, Pipe
• DriveWire SCF: scdwv, N, N1, N2, N3, N4, N5
• Storage (RBF): RBF, RBSuper, llwbsd, rbmem, DD, s0, s1, f0, f1, c0, c1, rbdw, X0, X1, X2, X3, dwio

### Loadable Modules

• 16550 Serial: sc16550, t0_sc16550
• WizFi / Network: wz, wz0, wz1, wz2, wz3, dwio_wizfi

recipes/wildbits/wildbits.mak
```diff
@@ -80,10 +80,11 @@ ifeq ($(FM),1)
 LFLAGS += -lfm
 CMDS_EXTRA += $(FM_CMDS)
 endif
+LOADMODS = $(SC16550) wz wz0 wz1 wz2 wz3 dwio_wizfi
 CMDS += $(STDCMDS) shell \
 	bootos9 scfg wbinfo wbreset modem \
 inetd telnet dw httpd $(BASIC09) $(BF) \
-	$(CMDS_EXTRA) wildspeed
+	$(CMDS_EXTRA) wildspeed $(LOADMODS)
```
DrPitre: Please explain what's going on here.

DrPitre (changes requested):

recipes/wildbits/wildbits.mak
rlucente-retro: The goal is to include multiple descriptors and modules in the disk image even if they don't fit  in the boot file. LOADMODS is intended to capture additional descriptors and modules to be added to the CMDS directory even though they are not in the boot file.

rlucente-retro: Should that just be collapsed into CMDS_EXTRA?

rlucente-retro:
This is synced with upstream and all the recent changes. Not sure if the WIZFI_MODS is needed just yet as @Nitrobotics may not be ready for that change. The key is to stay below the 32,256 threshold so the system doesn't freeze. Latest updates with the WIZFI_MODS are:

  • Raw module size: 30,685 bytes ($77DD)
  • Padded bootfile size: 30,720 bytes ($7800)
  • Hard limit: 32,256 bytes ($7E00)
  • Headroom remaining: 1,536 bytes (6 full 256-byte pages below the ceiling)

rlucente-retro:
I'm closing this as its just a ticking time bomb to make arbitary decisions on what's in/out of a bootfile. Any future change could again exceed the 32,256 byte limit. I'll create a separate PR for simply failing the build and alerting the user if the bootfile is too big.

PR #402, opened 2026-08-30
Wb/w6100 ethernet

Nitrobotics:
Build cmds/w6100eth test program.
Requires K2 core V8_rc5 or higher.
Demonstrated that the program connects and gets a ping reply.


MatthewMassie (approved):
Approved.

PR #403, opened 2026-08-30
DriveWire hardening

Nitrobotics:
Update rbdw.asm
Update dwinit_wildbits_serial.asm
Update dwrite_wildbits_serial.asm
Update dwread_wildbits_serial.asm
Update wildbits-drivewire-hardening.md (the fix explained)

This is not a willy-nilly whack-a-mole patch.  It took several days for Claude and I to run tests until the long dsave /x1 /s1 sessions were totally error-free with no lockups or OS crashes.  Also works in turbo mode.  I don't think noise on the USB channels stands a chance.

Jr2/K2 "V8" core is recommended.

### DIFF
--- a/docs/wildbits-drivewire-hardening.md
+++ b/docs/wildbits-drivewire-hardening.md
@@ -1,0 +1,173 @@
+# WildBits DriveWire Serial Driver Hardening
+
+**Dates:** 2026-08-28 / 2026-08-29
+**Files changed:**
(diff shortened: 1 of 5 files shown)


DrPitre:
I looked over the code. The one thing that concerns me is the leakage of wildbits-specific code into the driver. I would challenge Claude to see if this code could not be pushed further down into `dwio`.

PR #404, opened 2026-08-30
Add initial l2win recipe files

Nitrobotics:
Special recipe for Windows + cygwin64 systems that resolves several painful issues that keep popping up in the other recipes and makefiles.


MatthewMassie (approved):
Anything to make life easier for rapid development. I have used Cygwin and it was just what I needed. Approved.

PR #409, opened 2026-08-31
Move to big endian sprite registers

Nitrobotics:
Matches Core2x / V8_RC7 and higher

### DIFF
--- a/defs/wildbits.d
+++ b/defs/wildbits.d
@@ -682,50 +682,74 @@ SPRITE_SIZE0        equ       $20       00 = 32x32 - 01 = 24x24 - 10 = 16x16 - 1
 SPRITE_SIZE1        equ       $40
 
 
+* Sprite attribute records (128 total, 8 bytes each) live in VICKY page
+* $C0 at page offsets $1300-$16FF (record n at $1300+8*n). The SPn_*
+* equates below assume that page mapped in MMU slot 7 ($E000 window,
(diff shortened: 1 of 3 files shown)


MatthewMassie:
Roger with sprites now working normally on latest for Jr2 and K2, I think these changes would only be required for Jr and K models. I do not want to break anything before Tandy Assembly. I'm running your updated wildbits_k2_6809_v8_rc7.bit

Nitrobotics:
All that's changed is the SP#_Ctrl registers which precisely match the K2 and Jr2 Core2x/V8_RC7, and were inserted into wildbits.d by Claude himself, and proven through numerous test commands.  Nothing in the repo currently uses these registers, and if your outside programs referenced the wildbits.d from this PR then there's nothing to worry about in either case.

> Roger with sprites now working normally on latest for Jr2 and K2, I think these changes would only be required for Jr and K models. I do not want to break anything before Tandy Assembly. I'm running your updated wildbits_k2_6809_v8_rc7.bit

Nitrobotics:
In other words, have you tested this PR against your sprite demos to see if they actually break.  Please test.

> Roger with sprites now working normally on latest for Jr2 and K2, I think these changes would only be required for Jr and K models. I do not want to break anything before Tandy Assembly. I'm running your updated wildbits_k2_6809_v8_rc7.bit

MatthewMassie (approved):
Tested. Approved.

PR #411, opened 2026-09-02
wildbits: kernel/driver fixes bundle (slot-2 map-window safety + interrupt hygiene)

Nitrobotics:
Squash of wb/proper_slot_preservation and wb/interrupts_hygiene onto main. Field-verified on the K2 (l2_wildbitsk2_wc4u): wiz4up completes, four remote shells, console proc/wbinfo with all shells up, /f0 and /f1 clean.

Files:

level1/wildbits/modules/vtio.asm
  All nine MMU map windows (character write, scroll, erase-line, delete-
  line, text LUT, palette, two bitmap register paths, bell) no longer park
  the saved slot value on the stack. It goes to the new statics byte
  V.MapSav before the block is mapped and is restored from there after;
  no stack traffic crosses a block change inside a window (BellTone's
  and SetBitmap's temporaries moved to the IRQ-masked DP scratch
  D.IRQTmp). Same code assembles for Level 1 (FEU, slot 7) and Level 2
  (slot 2). Header comment explains the fault. Module grows 3152 -> 3215.

defs/wildbits_vtio.d
  Add V.MapSav (one byte) ahead of V.Last for the saved map-window slot.

level2/modules/kernel/krn.asm (all IFNE wildbits)
  - Cold start masks all four interrupt groups then clears all four pending registers with absolute stores, before any driver installs: the FPGA controller's registers survive software resets.
  - After marking global memory used, also mark system pages $40-$5F used so F$SRqMem never grows the system map into slot 2, the drivers' map window (8K of system RAM; 12 bytes of code).
  - The CrashDump padding before the constant-page tail is now `fill 255,KrnTail-*` with KrnTail equ $0F7F: lwasm computes the count, and an overrun fails the assembly ("Invalid fill length") instead of silently moving the tail. Image stays exactly 4096 bytes.

level2/modules/clock.asm
  Init clears only its own stale SOF pending bit and unmasks SOF with a
  read-modify-write, instead of writing whole INT_PENDING/INT_MASK
  registers (which wiped mousedrv's arming, since clock Init runs at the
  first F$STime, after vtio/mousedrv). Comment block documents why.

docs/wildbits-mmu-slot-safety.md
  New: symptom, how the RAM dumps found it, root cause, the two-layer
  fix, field verification table, rules for driver authors, open items.

### DIFF
--- a/defs/wildbits_vtio.d
+++ b/defs/wildbits_vtio.d
@@ -179,6 +179,7 @@ V.FONTNAME          RMB      33
 *V.SOLTable         RMB      32
                     endc
 
+V.MapSav            RMB       1                   saved MMU map-window slot value (vtio: no stack use inside a mapped window)
 V.InBuf             RMB       KBufSz              the input buffer
 V.KSBuf             RMB       KBufSz
                     RMB       250-.
--- a/docs/wildbits-interrupt-hygiene.md
+++ b/docs/wildbits-interrupt-hygiene.md
@@ -1,0 +1,66 @@
+# WildBits Interrupt-Controller Hygiene (krn + clock)
+
+**Date:** 2026-09-01
+**Platforms:** F256 WildBits K2 and Jr2
+**Status:** field-verified on the K2 (rc8 typematic test matrix, 2026-09-01), shipped in `wb/fixes_bundle`
+
+---
+
+## The problem
+
+The FPGA interrupt controller (`IRQ_Controller_Jr.v`) has four groups,
+each with PENDING ($FE20+n), POLARITY, EDGE and MASK registers. Its reset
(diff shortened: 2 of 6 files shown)


defs/wildbits_vtio.d
```diff
@@ -182,6 +182,7 @@ V.FONTNAME          RMB      33
 V.InBuf             RMB       KBufSz              the input buffer
 V.KSBuf             RMB       KBufSz
                     RMB       250-.
```
DrPitre, line 185: I know it's a minor nit, but can line 185 be moved up above line 182? That way we keep the buffers together and smaller variables stay together.

DrPitre (changes requested):

DrPitre (approved):

PR #415, opened 2026-09-04
wildbits.d defs core-matched updates

Nitrobotics:
Add new W6100 Ethernet, SplashFlash defs
Update WizFi and MIDI defs

Based directly off of the core wiring, so these updates are not guesses.


MatthewMassie (approved):
Approved.

PR #416, opened 2026-09-06
Play command improvements

Nitrobotics:
Update play.asm to reflect new MIDI address names in wildbits.d

Balance the volumes out between different music formats and sound chips on both the Jr2 and K2.

Set (Jr2/K2) conditional assembly of CODEC configuration

Add wmset - live WM8776 codec register writer for balancing by ear

wmset <reg> <value> (hex) writes one 16-bit control word to the codec at CODEC.Base using the same busy-wait sequence as vtio's SendToCODEC, with UPDATE set so attenuation changes take effect immediately.  Lets the DAC (R03/R04) and headphone (R00/R01) attenuations be tuned on a running machine instead of one disk build per guess.  Added to the wildbits CMDS list so it ships on the disk.

play: Lyra track volume drives CC11 expression; constant note velocity

Lyra's $E0 event is a track volume (0-7), not a per-note articulation, so it now sends CC11 (expression) on the track's channel through LyraExprConv (0,32,40,50,64,80,101,127: silent, then -24..0 dB in even 4 dB steps on the GM volume curve) instead of scaling note velocity through a 90..127 table that spanned only 3.5 dB.  Every note-on/off goes out at NOTE_VELOCITY (100). LYRA_DEFAULT_LEVEL 5 (CC11 80) is applied to all channels after MidiVolAll and to tracks that never send a volume event.  Per-track CC11 is sent only on change.  CC7 stays at 127.  Machine-independent: absolute loudness is the codec's job.

vtio: K2 DAC attenuation $D7 (-20dB); wmset notes updated

K2 headphone-jack balance re-trimmed by ear after play moved Lyra volume onto CC11 expression, which left .mus a little ahead of .lyr: R03/R04 $DF -> $D7, headphones stay $79.  Jr2 block unchanged ($FD/$60).

play: Lyra levels 1..7 -> CC11 80..127; level 0 stays silent

The -24..0 dB spread made Lyra's soft levels vanish on the SAM2695.  Lyra's lower volumes are still meant to be heard, so LyraExprConv now runs 0,80,87,93,101,109,118,127 (levels 1..7 = -8..0 dB in even 1.33 dB steps on the GM curve).  LYRA_DEFAULT_EXPR follows level 6 -> 118.

### DIFF
--- a/level1/wildbits/cmds/play.as
+++ b/level1/wildbits/cmds/play.as
@@ -88,7 +88,7 @@ edition = 16
 
 * Here are some tweakable options
 DOHELP              set       1                   1 = include help info
-SID_MAX_VOL         equ       15                  (0 = silence, 15 = loud) for all SID Voices
+SID_MAX_VOL         equ       6                   (0 = silence, 15 = loud) for all SID Voices - halved 2026-08-28 to balance against the quieter SAM2695 (.lyr) analog path
 SID_V1_CR1          equ       %00010001           SID Voice 1 Gate
 SID_V2_CR1          equ       %00010001           SID Voice 2 Gate
 SID_V3_CR1          equ       %00010001           SID Voice 3 Gate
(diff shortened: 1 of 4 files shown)


DrPitre (approved):

MatthewMassie (approved):
Approved.

PR #417, opened 2026-09-06
Wb/w6100 improvements

MatthewMassie:
Added sys/w6100ipconfig file to set IP, Subnet, Gateway, Port, Server IP, Mac on w6100 commands. This prevents users from recompiling to change IPs. This also adds a file receiver over Ethernet. If no config file is present command uses default compiled in addresses.


Nitrobotics (approved):
Looks good, keep up the good work.

PR #421, opened 2026-09-08
Wildbits 3-page slot 7 protection for FLink

Nitrobotics:
wildbits (2026-09-06): the core keeps THREE pages fixed at the top of every map ($FD00-$FFFF:
* the kernel's shadow-RAM tail and two I/O pages), not the CoCo3's two ($FE00-$FFFF).  The start-slot
* count therefore allows for $300 bytes instead of $200: A = (A + 3) / 32, overflow-safe.  A module
* whose last block is used past offset $1D00 starts the top-down search one slot lower and is never
* mapped into slot 7, where those bytes would be hidden (a 7470-byte vs hung both machines; 7212
* bytes worked).  B, the run length the search must find, is unchanged.

Update flink.asm

### DIFF
--- a/level1/modules/kernel/flink.asm
+++ b/level1/modules/kernel/flink.asm
@@ -131,6 +131,29 @@ FLinkModule         ldd       MD$MPtr,x ; get module pointer
                     ldy       MD$MPDAT,x ; get module DAT image pointer
                     ldd       MD$MBSiz,x ; get block size
                     addd      #$1FFF    ; round it up
+                  IFNE    wildbits
+* wildbits (2026-09-06): the core keeps THREE pages fixed at the top of every map ($FD00-$FFFF:
+* the kernel's shadow-RAM tail and two I/O pages), not the CoCo3's two ($FE00-$FFFF), so the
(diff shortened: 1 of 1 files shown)


jfed6000:
Is this the same size as the old F$Link for Wildbits?  Does the kernel need any of the padding code adjusted?

Nitrobotics:
> Is this the same size as the old F$Link for Wildbits? Does the kernel need any of the padding code adjusted?

Yes, it pads the same, only we have 3 extra bytes of future opcode space vs. non Wildbits.

jfed6000 (approved):
Approved

DrPitre:
Wait.. I thought the 3 page fix was already in the kernel and committed. If so, what's this change for?

Nitrobotics:
> Is this the same size as the old F$Link for Wildbits? Does the kernel need any of the padding code adjusted?


> Wait.. I thought the 3 page fix was already in the kernel and committed. If so, what's this change for?

Never was a PR or approved until now, but it's been in use on my feature overlay builds.  Works great.

PR #423, opened 2026-09-10
Updates to Play, Vs, Vtio codec

Nitrobotics:
The Play command got ahead of the new defs for it which are on this branch.
Small edit to vtio pertaining to the codec chip.
Vs command and defs.


MatthewMassie (approved):
Approved.

PR #425, opened 2026-09-13
Wb/1 mb pixview shellbg

MatthewMassie:
- Fix shellbg for 1MB memory allocation for bitmaps.
- Fix pixview for 1MB memory allocation for bitmaps.
- Added black CLUT for use with fadein/out.
- Build hexed (Hex editor) and pixview (Raw pixmap viewer) by default for L2.


Nitrobotics (approved):
Everything checks out.

PR #428, opened 2026-09-14
Wildbits: Max Ram Upgrade

Nitrobotics:
Wildbits krnp2: invoke the NoFlash map only if the core supports it.  Does not require changes to FEU. This is a L2 mode.
FEU/L1 = Flash always works, /f# and /c# drives.
L2 = Flash is mapped out exposing 768KB of unused RAM

Krnp2: MMU_IO_CTRL ($FFA1) bit 7 reads 1 on a core that implements FLASHDIS (rc16 and later).  krnp2 tests it once at boot, before anything is forked: 

 IF SET, it sets bit 2 (read-modify-write) and leaves LUT blocks $40-$9F out of the NotRAM gap table, so the flash/expansion window becomes 768 KB of pool RAM (K2: mfree 1696K);

 IF CLEAR, the gap table walked is the 1 MB map's ($40,$60 / $C0,$10 / $F0,$10) and $FFA1 is not written. wildbits.d: FLASHDIS / FLASHDIS.OK equates and the $FFA1 bit notes.


MatthewMassie (approved):
Tested.

